PR #13, opened 2025-09-25
add git precommit formatter for ocaml

ramadhanafif:
- **chore(nix): switch to mkShell and add dune_3 and ocamlformat**
- **chore(githooks): add ocamlformat pre-commit hook and update shell.nix**

added ocamlformat
added dune in nix
update shell.nix to copy githook when it starts


klarasm:
Looks good!

PR #14, opened 2025-09-25
ci: add build hook

klarasm:
I'm not sure this is entirely correct, but I cannot test on my computer.


klarasm:
Should ideally use nix for this part

PR #69, opened 2025-09-25
examples: add an SDF ForSyDe model and C threaded Linux impelementation

klarasm:


SamuelMiksits:
`examples/SDF_test.hs`:

It is good that we have examples, but do we want our compiler to accept this type of format (so everything is in the netlist, with actor specifications and function definitions inlined as lambdas), or do we want to have other programmatic restrictions? If we have it in `/examples` then by definition it must be usable with the tool, otherwise it is not an example. We will probably have some other `SDF_test.hs` so should it have a "numbered name" (like `SDF_test_1.hs`, or even `SDF_example_1.hs`)?

`examples/sdftest.c`:

Do we want to keep it somewhere else than the `/examples` base level folder, since it is not something that will be visualized or compilable with our tool? Another thing is we might want to write a GitHub action that pulls everything from the `/examples` folder and tries to compile it (not really a technical issue, can just pull `/examples/*.hs`)

klarasm:
Good points!

I'll rework the ForSyDe model into something more in line with what we have discussed.

I could perhaps also add some example input to both the model and the implementation (I'd just add a larger FIFO for input and prepopulate it from main, similar for output but print from main with same format as the haskell output).

I wasn't really sure were to put everything, didn't fit with what we have so far.

ramadhanafif:
I will accept this one first, @klarasm can update it later. The current one seems good enough as an example.

PR #70, opened 2025-09-26
docs: add grammar syntax definition

Himawari722:
Hi,

I have created a fundamental version of the syntax definition. I refered to Klara's SDF example code in Discord, so the syntax can support a more Haskell-like feature, like the lambda functions. Also the syntax will not differenciate the netlists, processes and functions and all treated as declarations. But this kind of high freedom of syntax might result in a difficulty in pattern matching. For the netlist we can specify that a function named system is necessary, just like the main function in C. 

The syntax might not be perfect and we can make it more specific later for easier implementation

### DIFF
--- a/docs/grammar.md
+++ b/docs/grammar.md
@@ -2,12 +2,32 @@ Grammar
 ========
 
 # Grammar Syntax Definition
-## Netlists
+```
+module                                                  (can ignore that)
+import                                                  (can ignore that)
+program -> {decl}                 
+decl    -> ident :: type                                (function declaration, type signature, get rid of "context" from Haskell)
+         | decllhs "=" declrhs                          (function definition)
+types   -> type
+         | types "->" types                             (function type)
+         | "("types {"," types} ")"                     (tuple type, at least 2 elements)
+         | "[" types "]"                                (list type)
+type    -> "Int" | "Signal" type                        (basic type or Signal wrapper)
+declrhs -> expr ["where" {decl}]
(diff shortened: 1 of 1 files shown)


SamuelMiksits:
It looks great so far! I think there was some discussion yesterday that we wanted to use expanded words and not as many abbreviations (such as `expression` instead of `expr`), but when looking at the grammar docs, they are empty for now. So maybe we need to implement that in the grammar docs first before we can finish this.

For now, can we either merge this to another branch on this repo that other people can modify, or can some other people become collaborators on your fork so we can update it after we have finished grammar docs?

EDIT: Sorry, I am a bit slow. You are updating the grammar docs, but maybe we wanted to have slightly different naming conventions.

Himawari722:
I’m a little bit confused about the workflow here: in the contribution doc it says we need to implement a feature in our own fork and send a PR to the main repo after finish? So the fork is kind of like an own thing? Or we implement features together in the fork of the people that is doing it?

klarasm:
If you're working independently it should be on your own fork.

If you want to, for shared work we can use a shared branch on Sebastians repository for easier access. I think we already have permission to create branches on Sebastians repo, perhaps we should call them something like 'collab/feature'.
You could also add the one you're collaborating with to your repository so you can work together it that way, or sharing the screen on discord or something else.

I'll update the contribution docs to make this more clear. Let me know if you have questions.

klarasm:
I'll merge this and we can change it later.

PR #78, opened 2025-09-30
parser: add the minimum version which can build

klarasm:


SamuelMiksits (approved):
Looks good! Now we actually have the overall structure and we have the necessary dependencies for menhir/ocamllex setup so that we can work on the compiler in the future

SamuelMiksits (changes requested):
It may be a good idea to add a comment about the `menhir --explain` flag. This flag will put a log file explaining potential shift reduce conflicts or other issues, by default the file for shift reduce conflicts will be placed in:
```
_build/default/lib/parser.conflicts
```

PR #79, opened 2025-10-01
Add example from IL2212 Lab as SDF_example_002 and add process constructor type signature for SDF_example_001 

SamuelMiksits:


klarasm (approved):
Looks good, now we have examples which togehter contains all of the basic actors (11, 12, 21, 22, delay).

PR #82, opened 2025-10-01
docs: update grammar doc with specifications

Himawari722:
Hi all, 

I have modified the naming of grammars avoiding using abbreviations. Hope that does not seem too messy.
I also added some specification on the Haskell and ForSyDe grammar that we currently support.
For the lambda functions, I think we can delete it since we don't really need to support that Haskell feature?

### DIFF
--- a/docs/grammar.md
+++ b/docs/grammar.md
@@ -1,33 +1,62 @@
 Grammar
 ========
 
+## Supported Haskell Features
+We have simplified the Haskell language as a small core subset for our current implementation. In particular, the features we supported are: 
+- **Declarations** 
+    - Function declarations with explicit type signatures (```f :: Int -> Int```)
+    - Function definitions (```f x = x + 1```)
+    - Value bindings (```x=5```)
+- **Simple Pattern Matching**
+    - On function arguments (simple patterns in ```f (x,y)=...```)
+    - On tuples (```(x,y)```patterns, at least 2 elements)
+    - On lists (```[x,y,z]```, list must have same element type)
+    - On simple literals (```x```, ```5```)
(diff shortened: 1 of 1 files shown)


klarasm:
We have worked on this in person where we iterated from yours and decided to have the grammar specification in the overleaf (same as we used for the research methodology course for now).

It's partly on us that we didn't inform you of this, but it is easier to communicate if you also come to meetings.

Himawari722:
> We have worked on this in person where we iterated from yours and decided to have the grammar specification in the overleaf (same as we used for the research methodology course for now).
> 
> It's partly on us that we didn't inform you of this, but it is easier to communicate if you also come to meetings.

Thank you for your information! I apologize that I will not be in Stockholm this week so I cannot  meet in person. I wonder if there is something than I can work on on my own?

klarasm:
Researching the static scheduling would be good to get started on and is probably a good fit for working on independently. You can assign yourself to the issue so it's tracked on github. The first step would probably be to begin writing some documentation for it.

Himawari722:
> Researching the static scheduling would be good to get started on and is probably a good fit for working on independently. You can assign yourself to the issue so it's tracked on github. The first step would probably be to begin writing some documentation for it.

Thank you! Will work on that

sthaeron:
After reviewing your work, @Himawari722, I found that having an introduction to the subset of Haskell and ForSyDe being used in the project is a great idea. I'll merge your request and update it slightly to better align with the specifications we wrote. I'll also add a reference to the specification at the end, mentioning that the specification document defines the syntax and semantics of the introduced subset.

PR #89, opened 2025-10-01
add docs for VS Code extension

ramadhanafif:


klarasm (approved):
Looks good!

sthaeron (commented):
Seems like a good starting point for the visualisation docs. For the issue #48 to be fully completed, it might be good to add how to interface with the vs code extension. Also might want to just have the vscode doc as a subsection of the visualisation docs.

sthaeron:
I'll merge this for now, we can combine visualisation and VSCode docs at a later stage

ramadhanafif:
I suppose that this vs code docs are going to be much longer than this, and more will be added as we found out new wonderful things for this lovable Javascript project. 

That is why I split it up from the visualization docs. Also, since VSCode is one of our project objective, i think it makes sense to have a dedicated docs for it.

PR #131, opened 2025-10-06
feat: Argument parsing for Haskell

### LINKED ISSUE #122: Compiler argument parsing
Update compiler to start argument parsing. Preferably use a argument parsing library to ensure scalibility of arguments. Start by being able to take an input file path as an argument. Relates to #121.

SamuelMiksits:
I put some time into this now because reading input files is highly blocking. Since we don't really have a nice Haskell project set up I am marking this as draft (might not be possible to merge depending on how we want to create the Haskell project, but the code is still important to have somewhere)

### DIFF
--- a/forsyde-devtools.cabal
+++ b/forsyde-devtools.cabal
@@ -10,7 +10,6 @@ license-file:       LICENSE
 -- copyright:
 build-type:         Simple
 extra-doc-files:    CHANGELOG.md
--- extra-source-files:
 
 common warnings
     ghc-options: -Wall
@@ -21,8 +20,11 @@ executable forsyde-devtools-exe
     hs-source-dirs:   src
     default-language: Haskell2010
     ghc-options:    -threaded -rtsopts -with-rtsopts=-N
+    other-modules:
+        Arguments
     build-depends:
         base >=4.6 && <6,
         ghc == 9.8.4,
         ghc-paths == 0.1.0.12,
-        forsyde-shallow == 3.5.0.0
+        forsyde-shallow == 3.5.0.0,
+        optparse-applicative >= 0.18.1.0,
--- a/src/Main.hs
+++ b/src/Main.hs
@@ -1,37 +1,116 @@
 module Main where
 
+import Arguments
 import GHC
-import GHC.Driver.Ppr
-import GHC.Paths
-import GHC.Utils.Outputable
+import GHC.Data.EnumSet (fromList)
 import GHC.Driver.Monad
-import GHC.Data.EnumSet ( fromList )
+import GHC.Driver.Ppr
+import GHC.Driver.Session (defaultFatalMessager, defaultFlushOut)
+import GHC.Paths
+import GHC.Utils.Outputable hiding ((<>))
+import Options.Applicative
 
-import GHC.Driver.Session ( defaultFatalMessager, defaultFlushOut )
+{-
 
-main = 
-    defaultErrorHandler defaultFatalMessager defaultFlushOut $ do
-      coreIR <- runGhc (Just libdir) $ do
-        dflags <- getSessionDynFlags
-        setSessionDynFlags dflags { 
-            ghcLink = LinkInMemory,
+    Specification
+
+    Compiler steps:
+
+    ForSyDe model ->
+    Core ->
+    ForSyDe IR ->
+    Procedural IR ->
+    C
+
+    - First we need a source file.
+    - Input mode:
+        - This should be the format for the input file, if not otherwise
+          specified, this should be a ForSyDe model (.hs) file
+
+    - Output mode:
+        - This should be the format of the output file, if not otherwise
+          specified, this should be C code
+        - if --output-core, output should be in core format
+        - if --output-ir-forsyde, output should be in ForSyDe-IR format
+        - if --output-ir-procedural, output should be in procedural IR format
+
+    - Output file argument:
+        - Use "-o" flag, outputs to desired file
+        - Dump to stdout if unspecified
+
+-}
+
+toCore :: FilePath -> IO CoreModule
+toCore input_file =
+  defaultErrorHandler defaultFatalMessager defaultFlushOut $ do
+    runGhc (Just libdir) $ do
+      dflags <- getSessionDynFlags
+      setSessionDynFlags
+        dflags
+          { ghcLink = LinkInMemory,
             ghcMode = CompManager,
-            generalFlags = fromList [
-                Opt_SuppressTicks,
-                Opt_SuppressCoercions,
-                Opt_SuppressCoercionTypes,
-                Opt_SuppressVarKinds,
-                Opt_SuppressModulePrefixes,
-                Opt_SuppressTypeApplications,
-                Opt_SuppressIdInfo,
-                Opt_SuppressUnfoldings,
-                Opt_SuppressTypeSignatures,
-                Opt_SuppressUniques,
-                Opt_SuppressStgExts,
-                Opt_SuppressStgReps,
-                Opt_SuppressTimestamps,
-                Opt_SuppressCoreSizes
-            ]
-        }
-        compileToCoreModule "examples/model/SDF_example_002.hs"
-      putStrLn $ showSDocUnsafe $ ppr coreIR
+            generalFlags =
+              fromList
+                [ Opt_SuppressTicks,
+                  Opt_SuppressCoercions,
+                  Opt_SuppressCoercionTypes,
+                  Opt_SuppressVarKinds,
+                  Opt_SuppressModulePrefixes,
+                  Opt_SuppressTypeApplications,
+                  Opt_SuppressIdInfo,
+                  Opt_SuppressUnfoldings,
+                  Opt_SuppressTypeSignatures,
+                  Opt_SuppressUniques,
+                  Opt_SuppressStgExts,
+                  Opt_SuppressStgReps,
+                  Opt_SuppressTimestamps,
+                  Opt_SuppressCoreSizes
+                ]
+          }
+      compileToCoreModule input_file
+
+write_output :: Output -> OutputFormat -> [Char] -> IO ()
+write_output StdOut _ s =
+  putStrLn s
+write_output (OutputFile (output_file, WithFileExtension)) _ s =
+  writeFile output_file s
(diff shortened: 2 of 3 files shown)


SamuelMiksits:
Potential feature improvements:
- You *have to* pass `-i` or `--input` for an input file. But this is pretty annoying for a compiler. It should just understand what file is the input file. Figure out how this can be done.
- By default it dumps the code to `stdout` if no output file is specified. I think it is fair enough if you pick some esoteric output format that it dumps to `stdout`, but for a regular end user they should be able to do say
`forsyde-devtools-exe ./Main.hs`
and it should just compile to a file called `main.c` or something.

Do we have something like Haddock set up? I think you need to do something like
```hs
--  |start of annotated comment
--  next line of annotated comment
```
or
```hs
{-|
  start of annotated comment
  next line of annotated comment
-}
```
instead of
```hs
-- start of comment that will not be annotated in Haddock
-- next line of comment that does not work properly
```
to [get comments to show up in it.](https://haskell-haddock.readthedocs.io/latest/markup.html#documenting-a-top-level-declaration) I think it is a good idea to figure this out now instead of later.

klarasm:
I think it would make sense to have a default output file (just like gcc defaults to a.out). Perhaps we could base it on the input file name and just change the extension?

For the input, I don't think we should just read some random file (I guess this should be handled by the user's build system). I could see that specifying no input file would default to standard in.

SamuelMiksits:
> I think it would make sense to have a default output file (just like gcc defaults to a.out). Perhaps we could base it on the input file name and just change the extension?

Yeah one of the reasons why I didn't make a default file is that the extension is context dependent (depends on flag). So for instance, the Core is like `.hcr` or something, and then for C we would probably want `main.c`. But actually this shouldn't be that big of a technical problem at all, you can remake the data structure so that you have a default name without a extension (like just  `main` and then just pattern match with say

```haskell
generate_default_file_name :: FilePath -> OutputFormat -> FilePath
generate_default_file_name o_f OutputC = o_f ++ ".c"
generate_default_file_name o_f OutputCore = o_f ++ ".hcr"
generate_default_file_name o_f OutputIRForSyDe = o_f ++ ".irf"
generate_default_file_name o_f OutputIRProcedural = o_f ++ ".irp"
```

or something (I don't know off the top of my head if `FilePath` can be concatenated like that), and then have `--stdout` as a flag that you use to put to `stdout`, and if you specify file name it doesn't add any extension

> For the input, I don't think we should just read some random file (I guess this should be handled by the user's build system). I could see that specifying no input file would default to standard in.

Any specific reason? The only `stdin` scenario I can see is if you do pipe cat into our program at which point reading from file makes more sense. Or should we be able to run the compiler from a REPL?

klarasm:
> > For the input, I don't think we should just read some random file (I guess this should be handled by the user's build system). I could see that specifying no input file would default to standard in.
> 
> Any specific reason? The only `stdin` scenario I can see is if you do pipe cat into our program at which point reading from file makes more sense. Or should we be able to run the compiler from a REPL?

Not really, I guess. I think for me it's just following with the convention but I don't have a strong opinion. I do, however think that having an option to read from stdin can be useful (e.g. if you have a tool which generates forsyde sdf models), but it doesn't have to be default, perhaps an additional flag `--stdin`?

SamuelMiksits:
Okay sure, `--stdin` flag doesn't sound too bad.

src/Main.hs
```diff
@@ -1,37 +1,154 @@
...
+        - if --output-ir-forsyde, output should be in ForSyDe-IR format
+        - if --output-ir-procedural, output should be in procedural IR format
+
+    - Output file argument:
+        - Use "-o" flag, outputs to desired file
+        - Dump to stdout if unspecified
+
+-}
+
+to_core_for_real i_f =
```
sthaeron: Haskell convention is to use camel case for functions, funny name though.

sthaeron (changes requested):
I managed to test your pull request locally using the github cli. Everything seems to be working well.

The `input-*` flags are unnecessary as we will not be making a lexer and parser for each layer of the compiler. Instead, when testing, we will define examples in Haskell code rather than from an input file. Having multiple output options is great as we will have pretty printing for each layer of the compiler.

sthaeron:
Is it possible to have the compiler compile to core by default? At least for now, since there is no need to have the option of outputting the Haskell code.

sthaeron:
> Do we have something like Haddock set up? I think you need to do something like

It seems like `haddock` is included with the `ghc` by default, meaning we already have it in our Haskell dev env. Seems like it can generate html or a latex pdf.

SamuelMiksits:
> The `input-*` flags are unnecessary as we will not be making a lexer and parser for each layer of the compiler. Instead, when testing, we will define examples in Haskell code rather than from an input file. Having multiple output options is great as we will have pretty printing for each layer of the compiler.

Not really how the feature is intended to work. We can input Haskell right now without lexing/parsing, so the idea is that you could have something else (for instance you have a .hs file in the same folder with say `exampleIRForSyDE` which is an example instantiated data structure containing some example model in ForSyDe IR, and choosing this compiler path loads that. 

> Is it possible to have the compiler compile to core by default? At least for now, since there is no need to have the option of outputting the Haskell code.

I don't understand, there is no option to output Haskell code. You can output in

```haskell
data OutputFormat
  = OutputC
  | OutputCore
  | OutputIRForSyDe
  | OutputIRProcedural
```
so in C, Core, ForSyDe IR or Procedural IR. The default option is to output in C, and you just put `--output-core` to output to core. But yes it is of course possible to switch so default is output in Core

sthaeron:
> I don't understand, there is no option to output Haskell code. You can output in

If you have no output flag and only input a file, it just outputs the Haskell code with ` To core! To ForSyDe IR! To Procedural IR! To C!` at the end.

sthaeron:
> so the idea is that you could have something else (for instance you have a .hs file in the same folder with say `exampleIRForSyDE` which is an example instantiated data structure containing some example model in ForSyDe IR

Yes, this is what we will be doing for testing. But I don't see how we can input a file; the pretty print functions need a variable that's defined in a file, not the file itself. How would you extract the name of the predefined variable from the file to give it to the pretty print functions?

SamuelMiksits:
> > I don't understand, there is no option to output Haskell code. You can output in
> 
> If you have no output flag and only input a file, it just outputs the Haskell code with ` To core! To ForSyDe IR! To Procedural IR! To C!` at the end.

Yes okay so this default mode is running all of the "dummy functions"
```haskell
run :: Arguments -> IO ()
-- "Normal run"
run (Arguments (InputFile i_f) outputFile InputForSyDe OutputC) = do
  file_content <- readFile i_f
  let file_output = toC $ toProceduralIR $ toForSyDeIR $ toCore $ file_content
  write_output outputFile file_output
```
since it has input format ForSyDe and output format C but if you want to I can replace what this does with something else. But this function calls takes in a file, compiles it to core, then takes the output of that, and translates the core into ForSyDe IR, then it takes the output of that and translates into Procedural IR, then that output is turned into C, and then it writes to output.


>Yes, this is what we will be doing for testing. But I don't see how we can input a file; the pretty print functions need a variable that's defined in a file, not the file itself. How would you extract the name of the predefined variable from the file to give it to the pretty print functions?

In REPL you can do `:load filename.hs` which will load all the modules/variables inside another file. Maybe you can do something similar in Haskell code? since I mean it is pretty absurd already that we can run GHC inside Haskell code. But I can also see your point, maybe you want to hard code import some other module instead

forsyde-devtools.cabal
```diff
@@ -25,4 +25,5 @@ executable forsyde-devtools-exe
         base >=4.6 && <6,
         ghc == 9.8.4,
         ghc-paths == 0.1.0.12,
-        forsyde-shallow == 3.5.0.0
+        forsyde-shallow == 3.5.0.0,
+        optparse-applicative >= 0.18.1.0
```
klarasm: Is it possible to leave a comma at the end here so we don't get unnecessary diffs if we add something else in the future?

SamuelMiksits:
> Okay sure, `--stdin` flag doesn't sound too bad.

Actually I kinda hate this idea now. Because `compileToCoreModule` has signature

```haskell
compileToCoreModule :: GhcMonad m => FilePath -> m CoreModule
```

which means you can't send a program to it as a string, you send a filepath for it to read from. There aren't other functions that allow you to send the program as raw text to I think I will not implement standard in

klarasm:
> > Okay sure, `--stdin` flag doesn't sound too bad.
> 
> Actually I kinda hate this idea now. Because `compileToCoreModule` has signature
> 
> ```haskell
> compileToCoreModule :: GhcMonad m => FilePath -> m CoreModule
> ```
> 
> which means you can't send a program to it as a string, you send a filepath for it to read from. There aren't other functions that allow you to send the program as raw text to I think I will not implement standard in

Ah, right, that's true. Feel free to ignore it in that case.

SamuelMiksits:
New push:
- My cabal was actually wrong. Now I fixed it so we have 
```
    other-modules:
        Arguments
```
which is how you properly make it so that if `Arguments.hs` gets modified the file gets recompiled (otherwise cabal returns "up to date" even if `Arguments.hs` has changed).
- I cleaned up a lot in general, now that I removed `InputFormat` it reduces the pattern matching matrix for main function. I still have the parsing and data type infrastructure in the `Arguments.hs` file but I removed the bloat in `Main.hs`
- You don't have to specify `-i`/`--input` anymore before your input file, because I figured out that you can use the `strArgument` function instead of say `str` to parse that without the flag.
- I implemented the naming schema that I mentioned before, so if you don't specify an output file the default behavior is now that it outputs to `main.xxx` where `xxx` depends on the output format, with the following file extensions depending on output formats:
```haskell
-- Function for generating default file names
generateDefaultFileName :: FilePath -> OutputFormat -> FilePath
generateDefaultFileName f OutputC = f ++ ".c"
generateDefaultFileName f OutputCore = f ++ ".hcr"
generateDefaultFileName f OutputIRForSyDe = f ++ ".irf"
generateDefaultFileName f OutputIRProcedural = f ++ ".irp"
```
and `--stdout` is a possible option to output to terminal instead of file.
- I made `OutputCore` the default options, so now there is no flag for `--output-core` and I added a flag for `--output-c` in the meanwhile

sthaeron (approved):
Tested different options and everything seems to be working!

Issue #106, opened 2025-10-06
Define Pico 2 board minimum runtime environment


SamuelMiksits:
We need to define how to use the Pico 2 with a minimum runtime environment. Does it use UART with printf/scanf, or do we want other hardware abstractions like buttons, switches and LEDs?
Part of #94

Issue #103, opened 2025-10-06
Rewrite circular buffers to they work on Pico 2


SamuelMiksits:
Create circular buffers that are compatible with the Pico 2 board.
Part of #94

Issue #98, opened 2025-10-06
Test manual C implementation on Pico 2


SamuelMiksits:
Just take some example C implementation of a SDF model and check that you can run it on the Pico 2 board.
Part of #94

Issue #97, opened 2025-10-06
Write docs on how to use Pico 2 board


SamuelMiksits:
Documentation on how to use Pico 2 board.
Part of #94

PR #136, opened 2025-10-07
Pico2 development environment

klarasm:
#98: For now it's easiest to create new project on ES-Lab-Kit, e.g. `newembproj -noRTOS SDF_example_002` and just link/copy the files to there.

#97: I added basic instructions for how use the board and bypass VSCode. I think bypassing VSCode should make it easier to integrate any local testing into the githooks (we could for example pass some example input to the UART and check the output).

#103: Just use the queues from the PicoSDK. These should also work seamlessly on multicore from my understanding. We then just need a partitioning and to schedule the actors for each separate core.

#106: Document the functions from the BSP. Also reference some of the documentation from PicoSDK which is likely to be useful for this project.

### DIFF
--- a/docs/es-labkit-board.md
+++ b/docs/es-labkit-board.md
@@ -119,3 +119,60 @@ From `bsp.h`.
 | `BSP_7SegDispFloat` | `void` | `float value` | Display floating `value` on the seven segment display |
 | `BSP_HasPSRam` | `size_t` | `void` | Get the size of the PSRAM (if any) |
 | `BSP_WaitClkCycles` | `void` | `uint32_t n` | Spin for `n` clock cycles (busy loop 3 cycles long) |
+
+### PicoSDK
+
+The information in this section is mostly just a summary of the [PicoSDK](https://www.raspberrypi.com/documentation/pico-sdk/index_doxygen.html#raspberry-pi-pico-sdk)
+documentation which is likely to be useful for this project.
+
+#### StdIO
+[`pico/stdio.h`](https://www.raspberrypi.com/documentation/pico-sdk/runtime.html#group_pico_stdio)
+
+Functions for more control over the IO on the pico2.
+Other than the standard input/output functions, it's also possible e.g. to limit IO
+to a specific driver.
+
+#### Time
+[`pico/time.h`](https://www.raspberrypi.com/documentation/pico-sdk/high_level.html#group_pico_time)
+
+Higher level abstractions around `hardware/timer.h`.
+
+This is divided into the modules:
+- [timestamp](https://www.raspberrypi.com/documentation/pico-sdk/high_level.html#group_timestamp):
+    Timestamp functions relating to points in time (including the current time).
+- [sleep](https://www.raspberrypi.com/documentation/pico-sdk/high_level.html#group_sleep):
+    Sleep functions for delaying execution in a lower power state.
+- [alarm](https://www.raspberrypi.com/documentation/pico-sdk/high_level.html#group_alarm):
+    Alarm functions for scheduling future execution.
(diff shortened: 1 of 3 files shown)


docs/es-labkit-board.md
```diff
@@ -0,0 +1,164 @@
+# Using the Embedded Systems Lab-Kit
+
+The [ES-LabKit](https://gits-15.sys.kth.se/mabecker/ES-Lab-Kit.git)
+is somewhat rigid in the way it wants to be installed.
+This makes sense as it is developed to be used in courses on school computers along with VSCode,
+but does make it a little harder to work with. Some of the scripts expect the ES-Lab-Kit
+repository to be cloned exactly into `~/Documents` and otherwise do not work without modification.
```
MichelD-Abo: The [install_student.sh](https://gits-15.sys.kth.se/mabecker/ES-Lab-Kit/blob/main/install_student.sh) is supposed to be only used on the lab computers. 
This link (https://gits-15.sys.kth.se/mabecker/ES-Lab-Kit?tab=readme-ov-file#getting-started-with-a-new-board-and-environment-on-your-own--computer) has the instructions for local installations. No need to clone exactly in `~/Documents`

docs/es-labkit-board.md
```diff
@@ -0,0 +1,164 @@
...
+your own cross-compiler, the triplet is `arm-none-eabi`.
+
+### GDB
+
+With default build options GDB only understands the host ISA,
+but it can be built to be multitarget.
+This might be a separate package depending on your distribution, in which case it
+will have to be installed separately.
+
+## Bypassing VSCode
```
MichelD-Abo, line 49: This [link](https://gits-15.sys.kth.se/mabecker/ES-Lab-Kit?tab=readme-ov-file#command-line-interface) shows also how build and flash the programs to the ES-Lab-Kit without the need of an IDE. You can link it here if you want.

MichelD-Abo (commented):
Looks good! Just added two comments.

docs/es-labkit-board.md
klarasm: For me the `newembproj` script complained about that. I did use the `install_student.sh` but modified it to change the path and leave out software I already had installed. I needed the following diff to make it work:
```
diff --git a/Tools/scripts/newembproj b/Tools/scripts/newembproj
old mode 100644
new mode 100755
index 0dc1e5714e3f..407b08e6abaa
--- a/Tools/scripts/newembproj
+++ b/Tools/scripts/newembproj
@@ -1,12 +1,12 @@
 #!/bin/bash
 
-if [[ ! -d ~/Documents/ES-Lab-Kit/Tools/ProjectCreator ]]
+if [[ ! -d ~/git/ES-Lab-Kit/Tools/ProjectCreator ]]
 then
   printf "Error: ProjectCreator not found here ~/Documents/ES-Lab-Kit/Tools/ProjectCreator. Did you run the install.sh script?\n"
   exit -1
 fi
 
-cd ~/Documents/ES-Lab-Kit/Tools/ProjectCreator
+cd ~/git/ES-Lab-Kit/Tools/ProjectCreator
 
 if [[ ! -d .venv ]]                                   # In case t he virtual environment does not exist...
 then
```

But I guess one could also use `ProjectCreator` directly

MichelD-Abo: Ok, this is weird. We always get new problems with the installations 🤣

klarasm: I updated the description to better match with this

docs/es-labkit-board.md
klarasm: Now added in addition to using GDB

MichelD-Abo (approved):
Looks good!

Issue #110, opened 2025-10-06
start scheduling docs


sthaeron:
Add a guide on how to use the methods created for single-core static scheduling, relates to #95.

Issue #102, opened 2025-10-06
Buffer sizes from schedule


sthaeron:
Define a method to obtain buffer sizes for signals from the schedule based on a repetition vector, relates to #95.

Issue #99, opened 2025-10-06
Repetition vector from topology matrix


sthaeron:
Define a method to obtain a repetition vector from a topology matrix, relates to #95.

Issue #96, opened 2025-10-06
Topology matrix from ForSyDe IR


sthaeron:
Define a function which creates a topology matrix based on the ForSyDe IR of a netlist, relates to #95.

PR #138, opened 2025-10-09
Static Scheduler

Himawari722:
Implemented a simple static scheduler. Added a scheduler doc (although I think this should be a part of c code generation? I did not find the c code generation doc issue). 

Relates to [#66] [#96] [#99] [#102] [#110].

There is a minor issue related to choosing edge to add initial tokens when there are cycles in the graph. I will create an issue on this and look into Ingo's lecture note or maybe just ask him.

Currently, the code has a simple representation of actor and edges, and the main function will print the result for testing. In the future, the input should be incorporated with ForSyDe IR and C generator. I will add this as a new issue


docs/scheduling.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Core Algorithm
+The algorithm works in three phases:
+1. **Topology Matrix and Repetition Vector Generation**
+    - Generate the Topology Matrix from the input actors and edges
+    - Calculate nullspace of the matrix and approximate to minimum integer
+    - Verify the approximated repetition vector
+2. **Schedule Generation and Initial Token Calculation**
+    - Choose the first fireable actor to fire
+    - When no actor is fireable, force fire the first one with remaining repetition (breaking deadlocks)
```
klarasm: Will this actually happen on a valid SDF graph? (i.e. there's no actor with enough input tokens)?

docs/scheduling.md
```diff
@@ -0,0 +1,30 @@
...
+    - Converts negative minimums to positive initial tokens
+3. **Buffer Analysis**
+    - Simulates the schedule execution
+    - Tracks maximum token usage for each edge
+    - Returns the maximum buffer size required
+
+## Limitation and Future Work
+Currently, this scheduler will generate a valid schedule, but not with a minimal buffer size. This could be optimized with new strategy of choosing fireable actors.
+
+In addition, when there are no fireable actors due to cycles in the graph, it will force fire the first actor with remaing repetition, and update its initial token count. For example, in SDF_example, there is an edge from C to D, and a delay edge from D to C. An initial token could be place on both edge to start a valid schedule, but the meaning and the functionality of the graph might differ. This should be check with Ingo.
```
klarasm: Isn't the delay actor just stating that this edge is where the initial token should be placed? I.e. if there is a delay actor between D and C, the edge (buffer) between D and C will have the initial token?

klarasm:
I'm not sure the scheduler should be placed under examples, but otherwise just quickly taking a glance this looks mostly good (I left two comments regarding the cycle and delay thing, but that might be me misunderstanding something). I'll try to run it later when I have time.

klarasm:
I run it and it looks correct for the SDF_examle_002 (which is the one replicated here, right?). I'm wondering if there's a way to keep the numbers rational throughout the process or is this a limitation of the library?

docs/scheduling.md
Himawari722: When I start implementing I did not notice that in ForSyDe delaySDF should be input with an initial token and my data are purely actors and edges, so I added this "force fire" to detect a required number. But with initial token specified in ForSyDe, we don't actually need this and I may change it later.

docs/scheduling.md
Himawari722: Yes, I misunderstood the placement of initial tokens and now I added the delay edge check.

Himawari722:
> I run it and it looks correct for the SDF_examle_002 (which is the one replicated here, right?). I'm wondering if there's a way to keep the numbers rational throughout the process or is this a limitation of the library?

To my current knowledge, yes, the most used linear algebra library use float on calculations. For a topology matrix for SDF graphs, I think this approximate calculation should be mostly correct.

Himawari722:
By the way, I just realized that in ForSyDe, and delay actor should have initial values lists:
d_1 :: Signal Int -> Signal Int
d_1 s = delaySDF [0] s 

and in the current ForSyDeIR, it seems like this is not included:
data IRConstructor
  = IRDelay String
  | IRActor String ActorType String
  
Do I misunderstand something? I think the inital values should be included and pass all the way down since c code also need this too. Could you have a look in this, @sthaeron ?

Himawari722:
I have updated the scheduler to support ForSyDeIR input. When trying to run the example system in ForSyDeIr.hs, I found out that was a problem of producing topology matrix for actors with self-loop and added checks for that.

There is another minor problem: for the identify actors that connects to an input, currently it simply finds an IRSignal that source is named "input". If there are multiple inputs, the name might be different and this would not work. There should be a new way to represent inputs in ForSyDeIR.

klarasm (approved):
Looks good (and well-commented)!

We still need to decide how to integrate this into the cabal project (do we want library, executable or something else?), but I think this is otherwise ready to be merged.

klarasm:
> I have updated the scheduler to support ForSyDeIR input. When trying to run the example system in ForSyDeIr.hs, I found out that was a problem of producing topology matrix for actors with self-loop and added checks for that.
> 
> There is another minor problem: for the identify actors that connects to an input, currently it simply finds an IRSignal that source is named "input". If there are multiple inputs, the name might be different and this would not work. There should be a new way to represent inputs in ForSyDeIR.

I think it migt be prudent to include the global input signals (and perhaps also global outputs) as a separate field in the `IRSystem` constructor (cc @sthaeron)?

Issue #105, opened 2025-10-06
Update contribution docs for Haskell


klarasm:
Update contribution docs for haskell, e.g. cabal instructions.

PR #141, opened 2025-10-10
Document testing and add example

klarasm:
#46, #47:
This adds testing with Hspec, using hspec-discover. Add testing documentation. Also add a simple test for the existing ForSyDeIR to have a starting point for future tests.

Note that this restructures the `forsyde-devtools.cabal` file quite a bit to avoid duplication. Let me know if I should change something here as I'm very much not an expert with cabal.

#105: I figured I could also resolve this, as I'm updating it for testing anyway.

#144: Now integrated into actions.

### DIFF
--- a/forsyde-devtools.cabal
+++ b/forsyde-devtools.cabal
@@ -11,19 +11,11 @@ license-file:       LICENSE
 build-type:         Simple
 extra-doc-files:    CHANGELOG.md
 
-common warnings
-    ghc-options: -Wall
-
-executable forsyde-devtools-exe
-    import:           warnings
-    main-is:          Main.hs
-    hs-source-dirs:   src
+common shared
+    ghc-options:      -Wall -threaded
     default-language: Haskell2010
-    ghc-options:    -threaded -rtsopts -with-rtsopts=-N
     other-modules:
-        Arguments
-        ForSyDeIR
-        CoreIR
+        Paths_forsyde_devtools
     build-depends:
         base >=4.6 && <6,
         ghc == 9.8.4,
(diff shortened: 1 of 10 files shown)


SamuelMiksits:
Ah, I updated the `forsyde-devtools.cabal` in [here](https://github.com/sthaeron/forsyde-devtools/commit/b9d2f34147e91c7885cd0423d9938e0ab5dcc773) so the merge is a bit spicy. I added the `,` at the end of the last line and it didn't give any trouble, so we don't need ugly diffs

klarasm:
Now updated with the fixed ForSyDeIR.

klarasm (commented):
Uh, forgot to stage the new test file in this commit, doh...

forsyde-devtools.cabal
```diff
@@ -11,23 +11,46 @@ license-file:       LICENSE
...
         forsyde-shallow == 3.5.0.0,
         optparse-applicative >= 0.18.1.0,
         filepath == 1.4.301.0,
+
+executable forsyde-devtools-exe
+    import:           shared
+    hs-source-dirs:   src
+    main-is:          Main.hs
+    ghc-options:      -rtsopts -with-rtsopts=-N
+    other-modules:
```
klarasm: The `other-modules` might not actually be needed as the executable now depends on the library

sthaeron (commented):
I found some issues when testing on my local repo. I managed to find fixes and clean up the nix and cabal configuration. Will be pushing the changes to this pull request. Otherwise everything else seemed to work great!

sthaeron:
Tested with a fresh install of cabal and managed to confirm testing packages are installed by nix, not cabal

sthaeron (approved):

PR #142, opened 2025-10-13
fix/ForSyDeIR

sthaeron:
Fixed IRDelay to include initial tokens as a list of ints. Was not able to correctly update the pretty printing of `IRFunction` expression, will need another update when we have our own pretty printing of the Core IR. Currently using `CoreBind` instead of `CoreExpr` as a temporary fallback.

### DIFF
--- a/docs/forsyde-ir.md
+++ b/docs/forsyde-ir.md
@@ -3,18 +3,20 @@ The ForSyDe IR can be found in the file `src/ForSyDeIR.hs`. The defined data typ
 ```haskell
 data ActorType
   = Actor11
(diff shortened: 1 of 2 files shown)


klarasm (approved):
Ran the `testForSyDeIR` function and it seems to work as expected, with the delay token.

klarasm:
@Himawari722 does this look good to you? I would like to merge this so I can fix my example for testing.

Himawari722:
Looks good to me, I will merge it

Himawari722 (approved):

PR #143, opened 2025-10-14
Feat/pretty print/core

MichelD-Abo:
Add custom pretty printing for Core and integrated it into Main.hs (arg parser)

### DIFF
--- a/src/Main.hs
+++ b/src/Main.hs
@@ -2,14 +2,6 @@ module Main where
 
 import Arguments
 import CoreIR (compileToCore, prettyCoreBindList)
-import GHC
-import GHC.Core (CoreProgram)
-import GHC.Data.EnumSet (fromList)
-import GHC.Driver.Monad
-import GHC.Driver.Ppr
-import GHC.Driver.Session (defaultFatalMessager, defaultFlushOut)
-import GHC.Paths
-import GHC.Utils.Outputable hiding ((<>))
 import Options.Applicative
 
 {-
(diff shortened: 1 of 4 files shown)


klarasm (approved):
Nice!
I pulled it and tested on SDF_example_002. It might be nice to have indentation for more than just the toplevel definitions/binds, but I don't think this is anything more than nice to have and should not block this.

src/Main.hs
SamuelMiksits, line 1: Can you have a second look on import list now that the code is moved to CoreIR.hs? I can build with only
```haskell
import Arguments
import CoreIR (compileToCore, prettyCoreBindList)
import Options.Applicative
```
so the rest generate a bunch of redundant import warnings

src/Main.hs
```diff
@@ -40,34 +42,33 @@ import Options.Applicative
 
 -}
 
-toCore :: FilePath -> IO CoreModule
```
SamuelMiksits, line 43: Comment out or just flat out remove?

src/Main.hs
```diff
@@ -100,9 +101,10 @@ run (Arguments (InputFile input_file) output_file OutputIRProcedural) =
   putStrLn "To Procedural IR"
 -- What we have so far, take input file and write out core
 run (Arguments (InputFile input_file) outputFile OutputCore) = do
```
SamuelMiksits: Okay so this is my bad from last time, but can you change 
```diff
+103 run (Arguments (InputFile input_file) outputFile OutputCore) = do
-103 run (Arguments (InputFile input_file) output_file OutputCore) = do
```
and then
```diff
-107 write_output outputFile OutputCore (prettyCoreBindList core)
+107 write_output output_file OutputCore (prettyCoreBindList core)
```
? The issue is that there is a function in my `Arguments.hs` that is called `outputFile` that gets shadowed, not that it is a big problem but it was something I missed so the naming is inconsistent with the other function calls. 

But that is also a larger question, do we want snake or camel for arguments also? It seems like most arguments I have done have been with snake and a lot of others are camel

SamuelMiksits (commented):
Looks cool. I agree with Klara a bit that the output can be hard to read with only 1 indent level, but I think that if you add tab to everything then we will pretty quickly tab into another zip code.

Here is example the last line of `NonRec(f_4:` in `example/model/SDF_example_002.hs`:
```
        Arg(Type(Int))))))))))), }))), }))), }))), }))), })))))))
```

which is 35 right parentheses/curls in a row. So even on the face of it we have already ran out of 80 cols even at 2 spaces per tab just alone on tabs before we even start pretty printing the line.

PR #154, opened 2025-10-20
wip: ProceduralIR definition and pretty-printing

Himawari722:
Created a draft version of Procedural IR based on Cigrid's AST and corresponding pretty-printing (not fully completed yet). Upload to see if this IR will be enough for converting ForSyDe IR to Procedural IR. 

During exploration implementing example AST according to the example c code, there are c language features that are not supported by Cigrid AST:

1. References. But maybe we can discard this and use everything in pointer
2. Cigrid only have SVarDef thus does not support "int i;", and I added SVarDecl since that is simple 
3. Cigrid does not support passing function as argument to another function. This can be discarded if each actor just call the origin name of those globally defined functions, which should be controlled during the conversion from ForSyDe IR to Procedural IR

### DIFF
--- a/src/ProceduralIR.hs
+++ b/src/ProceduralIR.hs
@@ -1,0 +1,222 @@
+module ProceduralIR where
+
+import GHC.Utils.Outputable
+
+data Uop
+
+data Bop
+
+-- T ::= TVoid | TInt | TChar | TIdent(r) | TPoint(T)
+data Type
+  = TVoid
+  | TInt
+  | TChar
+  | TIdent String
+  | TPoint Type
+
+-- Expressions
+-- e ::= EVar(r) | EInt(i) | EChar(c) | EString(r)
+--     | EBinOp(bop, e, e) | EUnOp(uop, e) | ECall(r, e)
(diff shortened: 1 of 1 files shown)


Himawari722:
I added some new grammar compared to Cigrid:

1. SFor: In Cigrid, for loops are "desugared" into SWhile. Since we are doing this reversly and translate AST to C code, SFor might be useful. The fields of SFor are:
`
SFor Statement Expression Statement Statement
`
which are initStatment (`int i=0`), condExpression(`i<n`), updateStatement(`++i`), and loop body. 

2. Added Increment(`++`) and Decrement(`--`) as unary opearators. Since associativity is not addressed in current AST, unary operators are all treated the same ways and placed at the left of the expression. This is currently fine since `++i` works in loop, but use with care if using this grammar in array access.

3. Btw, I named the AST of the operator as:
`
data Uop
= OPNegate
| OPLogicalNot
| OPIncrement
| OPDecrement
......
`
I don't know if this is good naming or the OP is redundant, can be changed in future

klarasm:
I'll merge this now for the reason mentioned on discord.

PR #159, opened 2025-10-24
Cleaning up pretty printing

sthaeron:
- Passing DynFlags to pretty printing functions for safe coversion of SDoc.
- Flags end in ir rather than start


klarasm (approved):
Nice to remove all the unnecessary GHC.(...) imports.

PR #161, opened 2025-10-25
example: Platform independent SDF_002 version with actor templates

SamuelMiksits:
# Create a platform independent implementation of `/examples/model/SDF_example_002.hs`

## Summary

This implementation should hopefully define the implementation details for how the backend of the compiler should translate into C code.

## Files

`actor_templates.h` - Header that defines actor templates. These are functions for `actorSDFXY`, that take the function that an actor should run, production/consumption rates, input/output buffers. Example:

```c
void actor11SDF(int consum, int prod,
                buffer_nonblocking* ch_in, buffer_nonblocking* ch_out,
                void (*f) (token*, token*));
```
Of note is that the operations are carried out on non-blocking buffers. Functions are defined for `actorSDF[1-4][1-4]`.

`actor_templates.c` - Implementations for above file, about 500 lines of code

`buffer_nonblocking.h` - Circular buffer implementation, platform agnostic in the sense that it uses `<stdlib.h>` `malloc`/`free`.

`buffer_blocking_pc.h` - `pthreads` implementation of blocking circular buffer.

`buffer_blocking_pico2.h` - Pico 2 board implementation of blocking read/write. Uses Pico 2 `bsp` (`<pico/util/queue.h>`)

`main_platform_independent.c` - Main file

`common.h` - This file acts as a wrapper for a lot of definitions. It pulls in the correct blocking buffer implementation depending on the platform. Internally this is implemented by "pre-processor enum defining" the platforms:

```C
/** @brief Selected target platform is a Pico 2 board */
#define PICO2 2

/**  @brief Platform modes, use this parameter to select platform.
 *   Available platforms so far: 
 *  @param PC - Linux PC (pthreads)
 *  @param PICO2 - PICO 2 board
 */
#ifndef PLATFORM
#define PLATFORM PC
#endif

#if PLATFORM == PC
    #include "buffer_blocking_pc.h"
#endif

#if PLATFORM == PICO2
    #include "buffer_blocking_pico2.h"
    #include <bsp.h>
#endif
```
### PC

Go to the folder `/examples/implementation/platform_independent/SDF_example_002/`

In order to run the PC version of the implementation, compile with

```
gcc ./main_platform_independent ./actor_templates.c -o main -DPLATFORM=PC
```
`-D` sets pre-processor definitions, and `-DPLATFORM=PC` is how common will be configured to use the PC implementation (this option is set by default), and `-DPLATFORM=PICO2` is how PICO2 will be chosen

Run with:
```
./main
```
Look at the file `./simulation.md`. It provides an example input sequence and the expected result. Confirm that this works as expected

### Pico 2

Follow the setup in `docs/es-labkit-board.md`.

Grab the contents from `/examples/implementation/platform_independent/SDF_example_002/` and move into your `ES-Lab-Kit/Software/example` folder or however you have chosen to create a project, or create your own project.

Modify the `CMakeLists.txt` so it has the following:

```cmake
add_executable(LabKitTest main_platform_independent.c ${BSP_SOURCES})
target_compile_options(LabKitTest PRIVATE -DPLATFORM=PICO2)
```
if the main binary is called `LabKitTest`

After building/running, open serial montitor. 

**NOTE: Make sure you have "Line ending" set to something else than "None". I have tested with "LF" and "CRLF", but "None" does not work**

As on PC, check `./simulation.md` and type the inputs and check that the result is correct.

**EDIT: I changed the CMake for Pico 2 as we have now removed the file `actor_templates.c`, the `add_executable` should now be:**

```cmake
add_executable(LabKitTest main_platform_independent.c ${BSP_SOURCES})
```


SamuelMiksits:
oops I have a mistake in `buffer_blocking_pc.h` on one line, will be force pushing soon

examples/implementation/platform_independent/SDF_example_002/actor_templates.c
```diff
@@ -0,0 +1,528 @@
+#include "actor_templates.h"
+
+void actor11SDF(int consum, int prod,
+                buffer_nonblocking* ch_in, buffer_nonblocking* ch_out,
```
klarasm: This is very nitpicky, but I prefer `buffer_nonblocking *ch_in` and so on. E.g. what does `int* a, b, c;` mean? Which of them are actually pointers?

Similar in a lot of other places which I'll not comment on further.

examples/implementation/platform_independent/SDF_example_002/actor_templates.c
```diff
@@ -0,0 +1,528 @@
...
+                void (*f) (token*, token*))
```
klarasm: My preference here would actually be `void (*f)(token *, token *)` (a space between type and asterisk) but this I don't care about as much.

examples/implementation/platform_independent/SDF_example_002/actor_templates.c
```diff
@@ -0,0 +1,528 @@
...
+{
+
+    token input[consum], output1[prod1], output2[prod2];
+    int i;
+
+    for(i = 0; i < consum; i++) {
+        read_token(ch_in, &input[i]);
+    }
+
+    f(input, output1, output2);
```
klarasm: Perhaps a blank line also after `f(...);`?

examples/implementation/platform_independent/SDF_example_002/actor_templates.c
```diff
@@ -0,0 +1,528 @@
...
+    }
+
+    for(i = 0; i < prod3; i++) {
+        write_token(ch_out3, output3[i]);
+    }
+
+    for(i = 0; i < prod4; i++) {
+        write_token(ch_out4, output4[i]);
+    }
+}
```
klarasm: No newline at the end of the file :(

examples/implementation/platform_independent/SDF_example_002/actor_templates.c
```diff
@@ -0,0 +1,528 @@
+#include "actor_templates.h"
```
klarasm: One thing to consider might also be to make all the actor functions static and have them in the header. This should give the compiler greater freedom to inline them, but also not include the versions unused in the binary since they aren't exported outside the file.

examples/implementation/platform_independent/SDF_example_002/actor_templates.h
```diff
@@ -0,0 +1,95 @@
+
+#ifndef ACTOR_TEMPLATES_H
+#define ACTOR_TEMPLATES_H
+
+#include "common.h"
+
+struct buffer_nonblocking;
+struct token;
```
klarasm: Should these really be in this file? Shouldn't these already be defined by the headers included in `common.h`?

examples/implementation/platform_independent/SDF_example_002/actor_templates.h
```diff
@@ -0,0 +1,95 @@
...
+                token*, token*, token*, token*, token*));
+
+void actor44SDF(int consum1, int consum2, int consum3, int consum4, int prod1,
+                int prod2, int prod3, int prod4, buffer_nonblocking* ch_in1,
+                buffer_nonblocking* ch_in2, buffer_nonblocking* ch_in3, buffer_nonblocking* ch_in4,
+                buffer_nonblocking* ch_out1, buffer_nonblocking* ch_out2, buffer_nonblocking* ch_out3,
+                buffer_nonblocking* ch_out4, void (*f) (token*, token*, token*, token*,
+                token*, token*, token*, token*));
+
+#endif
```
klarasm: No newline

examples/implementation/platform_independent/SDF_example_002/buffer_blocking_pico2.h
```diff
@@ -0,0 +1,37 @@
...
+{
+    queue_add_blocking(buf, &data);
+}
+
+static void read_token_blocking(buffer_blocking *buf, token *data)
+{
+    queue_remove_blocking(buf, data);
+}
+
+#endif
```
klarasm: No newline

examples/implementation/platform_independent/SDF_example_002/buffer_nonblocking.h
```diff
@@ -0,0 +1,71 @@
...
+
+static void read_token(buffer_nonblocking *buf, token *data)
+{
+	*data = buf->tokens[buf->tail];
+	buf->tail = (buf->tail + 1) %
+		buf->size;
+	--buf->used;
+}
+
+#endif
```
klarasm: No newline

examples/implementation/platform_independent/SDF_example_002/common.h
```diff
@@ -0,0 +1,41 @@
...
+#if PLATFORM == PC
+    #include "buffer_blocking_pc.h"
+#endif
+
+#if PLATFORM == PICO2
+    #include "buffer_blocking_pico2.h"
+    #include <bsp.h>
+#endif
+
+#endif
```
klarasm: No newline

examples/implementation/platform_independent/SDF_example_002/main_platform_independent.c
```diff
@@ -0,0 +1,150 @@
...
+
+    // Put initial token in channel a_4
+    write_token(a_4, 0);
+
+    // Repeating Schedule: A A B C D 
+    while(1) {
+        // Read input tokens
+        printf("Read 5 input tokens: ");
+        for(i = 0; i < 4; i++) {
+            scanf("%d", &input);
```
klarasm: Might be a good idea to check return of scanf (or feof(stdin) but nor sure if that's defined on the pico). What happens if you press CTRL-D on input?

examples/implementation/platform_independent/SDF_example_002/buffer_nonblocking.h
```diff
@@ -0,0 +1,71 @@
...
+ *  @param used - How namy elements are in use
+ *  @param head - Index for current position to write to (write before
+ *                increment)
+ *  @param tail - Index for current position to read from (read before
+ *                increment)
+ *  @param tokens - Array of elements
+ */
+typedef struct {
+	size_t size;
+	size_t used;
```
klarasm, line 19: This is never used besides assignment for this version. I guess we could output an error if there's a mismatch here, but otherwise we could just remove it.

examples/implementation/platform_independent/SDF_example_002/simulation.md
```diff
@@ -0,0 +1,20 @@
...
+3 35
+1 1 1 1 1 
+3 42
+0 0 0 0 0
+1 43
+0 0 0 0 0
+1 44
+1 5 2 3 4
+9 64
+```
```
klarasm, line 20: This works as expected for me on my laptop at least. Haven't bothered firing this up on the pico2 yet, though.

klarasm (changes requested):
I think this looks good overall but I have some nitpicks and other comments.

examples/implementation/platform_independent/SDF_example_002/actor_templates.c
klarasm: Also noticed some stray whitespace at otherwise empty lines in some of the files. (you can find them `grep -n ' $' ./*`)

examples/implementation/platform_independent/SDF_example_002/actor_templates.c
SamuelMiksits: The linker does not seem to be a fan
```
In file included from ./common.h:8,
                 from ./main_platform_independent.c:2:
./actor_templates.h:11:13: warning: ‘actor11SDF’ used but never defined
   11 | static void actor11SDF(int consum, int prod,
      |             ^~~~~~~~~~
./actor_templates.h:30:13: warning: ‘actor21SDF’ used but never defined
   30 | static void actor21SDF(int consum1, int consum2, int prod,
      |             ^~~~~~~~~~
./actor_templates.h:34:13: warning: ‘actor22SDF’ used but never defined
   34 | static void actor22SDF(int consum1, int consum2, int prod1, int prod2,
      |             ^~~~~~~~~~
/usr/bin/ld: /tmp/ccoHN8h7.o: in function `main':
main_platform_independent.c:(.text+0x633): undefined reference to `actor11SDF'
/usr/bin/ld: main_platform_independent.c:(.text+0x656): undefined reference to `actor11SDF'
/usr/bin/ld: main_platform_independent.c:(.text+0x679): undefined reference to `actor11SDF'
/usr/bin/ld: main_platform_independent.c:(.text+0x6ab): undefined reference to `actor21SDF'
/usr/bin/ld: main_platform_independent.c:(.text+0x6e5): undefined reference to `actor22SDF'
collect2: error: ld returned 1 exit status
```
I get this when I compile with gcc and have both files in one command. I tried compiling an obj and link afterwards but it gave me the same thing. Am I missing something obvious?

examples/implementation/platform_independent/SDF_example_002/actor_templates.h
SamuelMiksits: Actually it works fine without forward declaring them

examples/implementation/platform_independent/SDF_example_002/main_platform_independent.c
SamuelMiksits: Ctrl+D is not fun. It basically stops waiting for input at all and then just spins through the while loop like crazy. I think this is something worth looking into because in general it would be nice to do testing where you can do

```./main < test.in```

and then have a match on what the expected output is in a test case. And you know, the files usually end (a.k.a "end of file") so it would be nice if it stops gracefully in that case

File: examples/implementation/platform_independent/SDF_example_002/buffer_nonblocking.h (hunk 12 above)
Comment (SamuelMiksits) [reply examples/implementation/platform_independent/SDF_example_002/buffer_nonblocking.h:19]: Yeah I was thinking if you want to add something in the future if you wanted to check it in functions before calling those methods. But in general the whole point is that we should have buffer sizes pre-calculated so that we shouldn't run into buffer overruns.

File: examples/implementation/platform_independent/SDF_example_002/actor_templates.c (hunk 5 above)
Comment (klarasm) [reply examples/implementation/platform_independent/SDF_example_002/actor_templates.c]: Did you move the implementations from `actor_templates.c` or just mark the prototypes as static?

File: examples/implementation/platform_independent/SDF_example_002/actor_templates.c (hunk 4 above)
Comment (SamuelMiksits) [reply examples/implementation/platform_independent/SDF_example_002/actor_templates.c]: I think there were trailing whitespaces in almost every file. I found you can use Ctrl+K Ctrl+X in vscode to auto-remove them in the whole file so that's nice. I am a serial "trailing whitespacer", they just show up for some reason I don't know why

File: examples/implementation/platform_independent/SDF_example_002/actor_templates.c (hunk 5 above)
Comment (SamuelMiksits) [reply examples/implementation/platform_independent/SDF_example_002/actor_templates.c]: Ohhhh you wanted them in the header file. Okay

File: examples/implementation/platform_independent/SDF_example_002/actor_templates.c (hunk 4 above)
Comment (klarasm) [reply examples/implementation/platform_independent/SDF_example_002/actor_templates.c]: I think if you do a newline (which auto-indents) and then move from it in it does not automatically remove them in many editors or something.

File: examples/implementation/platform_independent/SDF_example_002/actor_templates.c (hunk 5 above)
Comment (SamuelMiksits) [reply examples/implementation/platform_independent/SDF_example_002/actor_templates.c]: Okay works now, do I remove `actor_templates.c`?

Comment (klarasm) [reply examples/implementation/platform_independent/SDF_example_002/actor_templates.c]: I think either way is fine, but as we likely won't include the header in any other file we won't get any duplicates in the binary and we might as well have them in the header and static.

File: examples/implementation/platform_independent/SDF_example_002/main_platform_independent.c (hunk 11 above)
Comment (klarasm) [reply examples/implementation/platform_independent/SDF_example_002/main_platform_independent.c]: Yep. If you check the return of `scanf` it will be zero (meaning it got no input). This might actually be a correctness issue as well as in that case we just reuse the old value. On the pico2, I think this is unlikely to be an issue, as that's reading from a uart and not a file or file descriptor.

File: examples/implementation/platform_independent/SDF_example_002/buffer_nonblocking.h (hunk 12 above)
Comment (klarasm) [reply examples/implementation/platform_independent/SDF_example_002/buffer_nonblocking.h:19]: We could perhaps wrap it in a debug ifdef or something. It might be nice in case we have issues in the static scheduler.

File: examples/implementation/platform_independent/SDF_example_002/actor_templates.c (hunk 5 above)
Comment (klarasm) [reply examples/implementation/platform_independent/SDF_example_002/actor_templates.c]: > Okay works now, do I remove `actor_templates.c`?

If you go with everything in the header, then yes.

**Comment by SamuelMiksits** [comment]
Okay I tried again :) I have not fixed Ctrl+D, but I should have fixed `(token*)` to `(token *)` and your `buffer_nonblocking *ch`, newlines and trailing whitespaces

File: examples/implementation/platform_independent/SDF_example_002/buffer_blocking_pico2.h [hunk 14]
```diff
@@ -0,0 +1,37 @@
+#ifndef BUFFER_BLOCKING_PICO2_H
+#define BUFFER_BLOCKING_PICO2_H
+
+#include <pico/util/queue.h>
+#include "bsp.h"
+
+// Wrap the pico2 implementation
+typedef queue_t buffer_blocking;
+typedef int token;
```
klarasm: You don't have this typedef in the headers for the PC. You defined the `typedef int token;` in `common.h` so you can probably remove it here as well.

examples/implementation/platform_independent/SDF_example_002/main_platform_independent.c
```diff
@@ -0,0 +1,150 @@
...
+
+actor_d :: Signal Int -> Signal Int -> (Signal Int, Signal Int)
+actor_d = actor22SDF (2,1) (1,2) f_4 where
+    f_4 [x, y] [z] = ([x + y + z], [x + y, x + y + z])
+*/
+
+
+// A = actor11SDF 2 1 f_a where
+//  f_a [x, y] = [x + y]
+void f_a(token* in, token* out) {
```
klarasm: These ones still have the asterisk attached to the type rather than the variable.

examples/implementation/platform_independent/SDF_example_002/common.h
```diff
@@ -0,0 +1,41 @@
+#ifndef COMMON_H
+#define COMMON_H
+
+// Types that we need
+typedef int token; // Temporary, needs to be generalized later.
```
klarasm: What you could do is do the typedef in the main file before you include this header. Then you wouldn't need to specify it in any of the header files at all.

examples/implementation/platform_independent/SDF_example_002/actor_templates.h
```diff
@@ -0,0 +1,554 @@
...
+{
+
+    token input[consum], output1[prod1], output2[prod2], output3[prod3];
+    int i;
+
+    for(i = 0; i < consum; i++) {
+        read_token(ch_in, &input[i]);
+    }
+
+    f(input, output1, output2, output3);
```
klarasm, line 60: You added a newline after some of the `f(...)` (or you already had them and I didn't notice) but missed it here.

examples/implementation/platform_independent/SDF_example_002/actor_templates.h
```diff
@@ -0,0 +1,554 @@
...
+    token input[consum], output1[prod1], output2[prod2], output3[prod3],
+        output4[prod4];
+
+    int i;
+
+    for(i = 0; i < consum; i++) {
+        read_token(ch_in, &input[i]);
+    }
+
+    f(input, output1, output2, output3, output4);
```
klarasm, line 92: Here too

examples/implementation/platform_independent/SDF_example_002/actor_templates.h
```diff
@@ -0,0 +1,554 @@
...
+
+    for(i = 0; i < consum1; i++) {
+        read_token(ch_in1, &input1[i]);
+    }
+
+    for(i = 0; i < consum2; i++) {
+        read_token(ch_in2, &input2[i]);
+    }
+
+    f(input1, input2, output);
```
klarasm, line 128: Here too

examples/implementation/platform_independent/SDF_example_002/main_platform_independent.c
klarasm: I was wrong, this also happens on the Pico2

klarasm (changes requested):
Mostly a few more nitpicks.

examples/implementation/platform_independent/SDF_example_002/common.h
SamuelMiksits: Good idea. Its probably better to have it there anyways because if you want to do say double you would also need to change scanf call so it would just be annoying to have scattered anyways

examples/implementation/platform_independent/SDF_example_002/actor_templates.h
SamuelMiksits: Do I remove all the 
```/* Read input tokens */```
```/* Write output tokens */```
as well? Only some have them

Comment (klarasm) [reply examples/implementation/platform_independent/SDF_example_002/actor_templates.h:60]: Yes please, didn't notice them. I don't really think they add anything here as the function/variable names are pretty self explanatory.

**Comment by klarasm** [comment]
Btw, took a look at the generated binary. If we ignore the f_* functions which are not marked as static, the only non-inlined function is `buffer_nonblocking_new` (the f_* functions are also inlined but still need to be present as they are not marked as static) and this is only at `-O2`. So the compiler even manages to inline the function pointers which is nice for us as we don't have to do that manually ourselves.

(It's a little funny, but if you make the f_* functions static, then `-O2` gives a smaller binary than `-Os`. Wait... This is even the case with the f_* as non-static)

File: examples/implementation/platform_independent/SDF_example_002/main_platform_independent.c (hunk 11 above)
Comment (SamuelMiksits) [reply examples/implementation/platform_independent/SDF_example_002/main_platform_independent.c]: Do you have any smarter idea than literally just putting
```c
int scanf_out = 0;
```
and then
```c
read_token(s_out, &output);
if (scanf_out == EOF)
    return 0;
```
on literally every scanf? It feels horrible but there is absolutely no way I am putting

```c
while(scanf_out != EOF) {
        // Read input tokens
        printf("Read 5 input tokens: ");
        for(i = 0; i < 4; i++) {
            scanf_out = scanf("%d", &input);
            if (scanf_out == EOF)
                break;
            write_token(s_in_a, input);
        }
        if (scanf_out == EOF)
            continue;
        for(i = 0; i < 1; i++ ) {
            scanf_out = scanf("%d", &input);
            if (scanf_out == EOF)
                break;
            write_token(s_in_b, input);
        }
        if (scanf_out == EOF)
            continue;
```

SamuelMiksits: Okay actually it seems like it works with only the `break` at least on PC

SamuelMiksits: Okay my bad, it actually doesn't. So we literally need either return or this abomination that I posted above

klarasm: You could put the scanf directly in the if(...) I guess. (I don't think you need the continues, as we would have already broken out of the loop in that case. edit: nvm, saw you already tried this)

klarasm: > Okay my bad, it actually doesn't. So we literally need either return or this abomination that I posted above

Ah, you're right. The first only breaks out of the inner for loops. You could maybe use goto (which can be fine if used correctly, e.g. in error handling). Another option could be to have the schedule in a separate (static) function, and we can use return which the compiler will basically turn into goto for us so we don't need to bother about using it correctly.

klarasm: The separate function is a little annoying as we have to pass a lot of arguments. This is how I would do the goto solution anyhow:
```diff
diff --git a/examples/implementation/platform_independent/SDF_example_002/main_platform_independent.c b/examples/implementation/platform_independent/SDF_example_002/main_platform_independent.c
index f158e670bd18..f9ec05546056 100644
--- a/examples/implementation/platform_independent/SDF_example_002/main_platform_independent.c
+++ b/examples/implementation/platform_independent/SDF_example_002/main_platform_independent.c
@@ -106,11 +106,13 @@ int main() {
         // Read input tokens
         printf("Read 5 input tokens: ");
         for(i = 0; i < 4; i++) {
-            scanf("%d", &input);
+            if (scanf("%d", &input) < 1)
+                   goto err_free;
             write_token(s_in_a, input);
         }
         for(i = 0; i < 1; i++ ) {
-            scanf("%d", &input);
+            if (scanf("%d", &input) < 1)
+                   goto err_free;
             write_token(s_in_b, input);
         }
 
@@ -138,6 +140,7 @@ int main() {
         printf("\n");
     }
 
+err_free:
     // Free all buffers
     buffer_nonblocking_free(s_in_a);
     buffer_nonblocking_free(s_in_b);
@@ -146,5 +149,7 @@ int main() {
     buffer_nonblocking_free(a_2);
     buffer_nonblocking_free(a_3);
     buffer_nonblocking_free(a_4);
+
+    printf("\nExiting due to no or bad input\n");
     return 0;
 }
```

klarasm: The generated code is a little less nice with this (probably due to the additional branches and `actor11SDF` is not inlined anymore.
Here is another version which is less bad for inlining:
```diff
diff --git a/examples/implementation/platform_independent/SDF_example_002/main_platform_independent.c b/examples/implementation/platform_independent/SDF_example_002/main_platform_independent.c
index f158e670bd18..b256158c5f28 100644
--- a/examples/implementation/platform_independent/SDF_example_002/main_platform_independent.c
+++ b/examples/implementation/platform_independent/SDF_example_002/main_platform_independent.c
@@ -78,7 +78,7 @@ int main() {
     token output;
 
     // C99 array iteration variable
-    int i;
+    int i, ret;
 
     //Create FIFO-Buffers for signals
 
@@ -106,11 +106,11 @@ int main() {
         // Read input tokens
         printf("Read 5 input tokens: ");
         for(i = 0; i < 4; i++) {
-            scanf("%d", &input);
+            ret = scanf("%d", &input);
             write_token(s_in_a, input);
         }
         for(i = 0; i < 1; i++ ) {
-            scanf("%d", &input);
+            ret = scanf("%d", &input);
             write_token(s_in_b, input);
         }
 
@@ -136,6 +136,9 @@ int main() {
             printf("%d ", output);
         }
         printf("\n");
+
+        if (ret < 1)
+            break;
     }
 
     // Free all buffers
```
(I know the first ret = ... is not necessary, but should be easier to generate if we don't have to care)

Here we will produce bad output for one schedule, but we will exit, and if we have the amount of input aligned to the schedule there will be no bad output.

SamuelMiksits: Okay we can do goto. I mean I am personally not super concerned about cleanup. Like it literally does not matter if you leak memory when you exit, the program is exiting and giving it all back anyways. It is worse if you are constantly mallocing new things and never giving it back during runtime.

Having schedule in separate function *might* make sense, because you would have to have fork/join in main for multi-threading and then the malloced non-blocking buffers and schedule in each function defined per thread, so that way of looking at the problem could potentially make mullti-threading easier to add later?

klarasm: Or even:
```diff
diff --git a/examples/implementation/platform_independent/SDF_example_002/main_platform_independent.c b/examples/implementation/platform_independent/SDF_example_002/main_platform_independent.c
index f158e670bd18..0c071627854b 100644
--- a/examples/implementation/platform_independent/SDF_example_002/main_platform_independent.c
+++ b/examples/implementation/platform_independent/SDF_example_002/main_platform_independent.c
@@ -78,7 +78,7 @@ int main() {
     token output;
 
     // C99 array iteration variable
-    int i;
+    int i, ret = 1;
 
     //Create FIFO-Buffers for signals
 
@@ -102,15 +102,15 @@ int main() {
     write_token(a_4, 0);
 
     // Repeating Schedule: A A B C D
-    while(1) {
+    while(ret == 1) {
         // Read input tokens
         printf("Read 5 input tokens: ");
         for(i = 0; i < 4; i++) {
-            scanf("%d", &input);
+            ret = scanf("%d", &input);
             write_token(s_in_a, input);
         }
         for(i = 0; i < 1; i++ ) {
-            scanf("%d", &input);
+            ret = scanf("%d", &input);
             write_token(s_in_b, input);
         }
 
```

klarasm: > Okay we can do goto. I mean I am personally not super concerned about cleanup. Like it literally does not matter if you leak memory when you exit, the program is exiting and giving it all back anyways. It is worse if you are constantly mallocing new things and never giving it back during runtime.
> 
> Having schedule in separate function _might_ make sense, because you would have to have fork/join in main for multi-threading and then the malloced non-blocking buffers and schedule in each function defined per thread, so that way of looking at the problem could potentially make mullti-threading easier to add later?

Hmm. This might be a good point. In the multi-threaded case we would also need to make the buffers global variables anyway, so we might as well do it now. But I think it still might make sense to only check the scanf return at the end (or beginning) of the loop due to optimization. I guess we have to decide how much we care about correct output.

SamuelMiksits: > Here we will produce bad output for one schedule, but we will exit, and if we have the amount of input aligned to the schedule there will be no bad output.

Yeah my main concern is period aligned input files. Or is that a reasonable limitation that we can enforce? If it is then I think whatever will have the best performance wins. I also feel like I am doing something wrong if I have to argue that a goto-based solution is "the most elegant" XD

klarasm: Another option to consider is that we could read all input for a schedule run in one go. Sure, this would take a little more memory, but will give us both correctness and performance.

But I think either way that splitting the schedule out of main is probably a good idea.

klarasm: (btw, my solution above does not actually handle period-aligned files. It will get the last input correctly, but then fail at the first input of the next (non-existing) period)

SamuelMiksits: I am still thinking that you have blocking read/write buffers global, but then you have everything (like malloc local non-blocking, initialize delay, loop with schedule, free local non-blocking) inside the function for that thread.

SamuelMiksits: Okay I think I will just use goto for now and do another force push that solves the other things as well. Then we can look at the rest some other time I will add a `test.in` file, unfortunately if I run
```
./main < test.in > test.out
```
the `test.out` file looks like this
```
Read 5 input tokens: Output: 3 11 
Read 5 input tokens: Output: 5 24 
Read 5 input tokens: Output: 3 35 
Read 5 input tokens: Output: 3 42 
Read 5 input tokens: Output: 1 43 
Read 5 input tokens: Output: 1 44 
Read 5 input tokens: Output: 9 64 
Read 5 input tokens: 
```
which is quite horrible. We can probably have a "clean IO" pre-processor flag with build system that either does/doesn't print the Read input tokens/Output things

SamuelMiksits:
(It took me like 3 force pushes to get there, I forgot to add the test.in file, then I had one function signature that was very long on one line, so now all lines in `actor_templates.h` should be sub 80 columns)

examples/implementation/platform_independent/SDF_example_002/main_platform_independent.c
klarasm: I don't really like the goto solution anymore due to the de-inlining side-effect. I would rather read all input in one go and then break right after before we run any actor. But I agree we can look at a better solution sometime else.

SamuelMiksits: There is a problem. So if I run your pthreads `SDF_example_001.c` I get the following:
```
2 2 1 1
Process B: 2
Process B: 0
Process C: 21
Process C: 4
2 2 1 1 
Process B: 2
Process C: 21
Process B: 0
Process C: 4
``` 
The implementation is not incorrect, but the input/output order is nondeterministic (in your example only one process is input, so the input in that example is fine) when you have multi-threading. If we want to handle all input/output at once, we would need another condvar basically that halts all actors at the start of a schedule until all input has been provided, and another condvar that blocks writing to output until all actors have provided their outputs for IO so the data can be created in the predetermined order.

So maybe the output approach in this example is better (since it prints which output the actor comes from), and if you write a test case for multi-threading the matcher needs to check that the output from within each actor is the same as the output from each actor in the test file since the output order within an actor should always be the same if input is properly synchronized. But this still requires the input synchronization condvar.

In IL2212 Lab 3 this problem was kind of obfuscated away because there was only one actor that received input and one actor that prints output, but this is also probably the most realistic scenario for a data flow application since it describes a pipeline that data flows through.

klarasm: I think this is also something which would not happen in a real usecase. The reason to have multiple global input/output signals would always(?) be that they are generated or consumed by a different entity, in which this would not be (as large of) a problem.

If we want deterministic order, it should be enough to have all input on one core/thread and output on one core/thread (though not necessarily the same). For testing, this is probably a good idea. We could also use additional synchronozation instead as you suggest (essentially an ordered barrier), but not sure if we have the right primitives for that on the pico2.

SamuelMiksits:
For Pico 2 file sizes. I haven't looked at the elf analyzer for how the program is optimized, but I am getting the following output files when running with `-g` and `-OX`: (`ls -alh` command`)

`-O2`:

```
.rwxr-xr-x. sam sam  80 KB Sun Oct 26 09:24:24 2025 LabKitTest.bin
.rw-r--r--. sam sam 802 KB Sun Oct 26 09:24:24 2025 LabKitTest.dis
.rwxr-xr-x. sam sam 726 KB Sun Oct 26 09:24:24 2025 LabKitTest.elf
.rw-r--r--. sam sam 531 KB Sun Oct 26 09:24:24 2025 LabKitTest.elf.map
.rw-r--r--. sam sam 225 KB Sun Oct 26 09:24:24 2025 LabKitTest.hex
.rw-r--r--. sam sam 161 KB Sun Oct 26 09:24:24 2025 LabKitTest.uf2
```

`-O0`:

```
.rwxr-xr-x. sam sam  96 KB Sun Oct 26 09:25:05 2025 LabKitTest.bin
.rw-r--r--. sam sam 1.1 MB Sun Oct 26 09:25:05 2025 LabKitTest.dis
.rwxr-xr-x. sam sam 508 KB Sun Oct 26 09:25:04 2025 LabKitTest.elf
.rw-r--r--. sam sam 578 KB Sun Oct 26 09:25:04 2025 LabKitTest.elf.map
.rw-r--r--. sam sam 269 KB Sun Oct 26 09:25:05 2025 LabKitTest.hex
.rw-r--r--. sam sam 192 KB Sun Oct 26 09:25:05 2025 LabKitTest.uf2
```

`-Os`:

```
.rwxr-xr-x. sam sam  76 KB Sun Oct 26 09:26:04 2025 LabKitTest.bin
.rw-r--r--. sam sam 740 KB Sun Oct 26 09:26:04 2025 LabKitTest.dis
.rwxr-xr-x. sam sam 607 KB Sun Oct 26 09:26:04 2025 LabKitTest.elf
.rw-r--r--. sam sam 538 KB Sun Oct 26 09:26:04 2025 LabKitTest.elf.map
.rw-r--r--. sam sam 214 KB Sun Oct 26 09:26:04 2025 LabKitTest.hex
.rw-r--r--. sam sam 152 KB Sun Oct 26 09:26:04 2025 LabKitTest.uf2
```

I find it amusing that the `.elf` is larger with `-Os` than `-O0`, but this is why I included the other file sizes as well since `.dis` is almost half the size in `-Os`. I have to be honest I have no idea what the different files are

examples/implementation/platform_independent/SDF_example_002/main_platform_independent.c
klarasm: I outlined how I would solve these problems for the pthreads version in #162.

klarasm: This would basically be:
```diff
diff --git a/examples/implementation/platform_independent/SDF_example_002/main_platform_independent.c b/examples/implementation/platform_independent/SDF_example_002/main_platform_independent.c
index bec989f251af..48cba1c4011b 100644
--- a/examples/implementation/platform_independent/SDF_example_002/main_platform_independent.c
+++ b/examples/implementation/platform_independent/SDF_example_002/main_platform_independent.c
@@ -76,7 +76,8 @@ int main() {
     #endif
 
     // Temporary tokens for print/scan functions for input and output
-    token input;
+    token input_a[4];
+    token input_b[1];
     token output;
 
     // C99 array iteration variable
@@ -105,18 +106,21 @@ int main() {
 
     // Repeating Schedule: A A B C D
     while(1) {
+        int ret;
         // Read input tokens
         printf("Read 5 input tokens: ");
-        for(i = 0; i < 4; i++) {
-            if (scanf("%d", &input) < 1)
-                goto err_free;
-            write_token(s_in_a, input);
-        }
-        for(i = 0; i < 1; i++ ) {
-            if (scanf("%d", &input) < 1)
-                goto err_free;
-            write_token(s_in_b, input);
-        }
+        for(i = 0; i < 4; i++)
+            ret = scanf("%d", &input_a[i]);
+        for(i = 0; i < 1; i++)
+            ret = scanf("%d", &input_b[i]);
+
+        if (ret < 1)
+            break;
+
+        for(i = 0; i < 4; i++)
+            write_token(s_in_a, input_a[i]);
+        for(i = 0; i < 1; i++)
+            write_token(s_in_b, input_b[i]);
 
         // A
         actor11SDF(2, 1, s_in_a, a_1, f_a);
@@ -142,7 +146,6 @@ int main() {
         printf("\n");
     }
 
-err_free:
     // Free all buffers
     buffer_nonblocking_free(s_in_a);
     buffer_nonblocking_free(s_in_b);
```

This is a little more involved but would allow us to avoid having too many branches in the loop.

examples/implementation/platform_independent/SDF_example_002/main_platform_independent.c
```diff
@@ -0,0 +1,155 @@
...
+    // Repeating Schedule: A A B C D
+    while(1) {
+        // Read input tokens
+        printf("Read 5 input tokens: ");
+        for(i = 0; i < 4; i++) {
+            if (scanf("%d", &input) < 1)
+                goto err_free;
+            write_token(s_in_a, input);
+        }
+        for(i = 0; i < 1; i++ ) {
```
klarasm: Btw, an extra space after `i++`

examples/implementation/platform_independent/SDF_example_002/main_platform_independent.c
```diff
@@ -0,0 +1,155 @@
...
+
+actor_d :: Signal Int -> Signal Int -> (Signal Int, Signal Int)
+actor_d = actor22SDF (2,1) (1,2) f_4 where
+    f_4 [x, y] [z] = ([x + y + z], [x + y, x + y + z])
+*/
+
+
+// A = actor11SDF 2 1 f_a where
+//  f_a [x, y] = [x + y]
+void f_a(token *in, token *out) {
```
klarasm: Maybe also make these `static`?

klarasm:
I think regarding placing the schedule in a separate function (so we are compatible with multicore  when we get there) you can just go for it.

For input/output to have deterministic order, I think it will be easiest to either:
- have both input and output on the same core/thread. This is a little overly constrained, but if this is easier to hook in the rest of the compiler then it's a valid option.
- have all input in the same core/thread, and all output in another core/thread. This will give the properties we want but be a little less constrained. It's still more constrained than is possible, but this should also be easy to make work with only the queues on the pico2.
- There are a few other synchronization primitives on the pico2 which I seem to have missed when doing the initial documentation (in pico_sync, e.g: `pico/mutex.h`, `pico/sem.h`. See https://www.raspberrypi.com/documentation/pico-sdk/high_level.html#group_pico_sync) we could look at if we want more fine-grained synchronization however.

examples/implementation/platform_independent/SDF_example_002/main_platform_independent.c
SamuelMiksits: Fixed

examples/implementation/platform_independent/SDF_example_002/main_platform_independent.c
SamuelMiksits: I made them static

examples/implementation/platform_independent/SDF_example_002/main_platform_independent.c
SamuelMiksits: Okay I am fine with this.

klarasm (approved):
I think this looks good now, thanks!

SamuelMiksits:
Marking as "ready for review" now lol

PR #162, opened 2025-10-28
Fix the pthreads implementation of SDF_example_001

klarasm:
This is how I would solve the EOF problem and deterministic output order for the pthreads variant (as described in #161).

Additionally I had a typo in `fifo_put_multiple()` which resulted in a deadlock which is also fixed here. But you didn't use the multiple element versions in #161 so you can probably ignore this.

For the pico2 implementation, this would essentially translate into:
1. At the start of the schedule, read all needed input to run this instance to completion
  - If we didn't get enough input, exit at this point
  - This will take a little more memory, but means we won't have too many exit points of the schedule loop, meaning the compiler will be more likely to keep inlining stuff.
2. For multicore, always place all input on the same core, and all output on the same core (in the case we want deterministic input and output order).

### DIFF
--- a/examples/implementation/SDF_example_001.c
+++ b/examples/implementation/SDF_example_001.c
@@ -139,28 +139,48 @@ static void fifo_get_multiple(struct fifo *fifo, size_t count, element elements[
 }
 
 enum process {
-	ProcessA = 0,
+	Input = 0,
+	ProcessA = 1,
 	ProcessB,
 	ProcessC,
 	ProcessD,
+	Output,
 	End
 };
 
+struct fifo *fifo_in_a;
 struct fifo *fifo_a_b;
 struct fifo *fifo_a_c;
 struct fifo *fifo_b_d;
 struct fifo *fifo_c_d;
+struct fifo *fifo_b_out1;
+struct fifo *fifo_d_out2;
+
+static void *process_input(void *arg)
+{
+	element input[2];
+	int ret;
+
+	fprintf(stderr, "Process Input starting\n");
+	while (!feof(stdin)) {
+		ret = scanf(E_F " " E_F, &input[0], &input[1]);
+		if (ret < 2)
+			break;
+		fifo_put_multiple(fifo_in_a, 2, input);
+	}
+	fprintf(stderr, "Input exiting\n");
+	pthread_exit(NULL);
+}
 
 static void *process_a(void *arg)
 {
-	element input;
+	element input[2];
 
 	fprintf(stderr, "Process A starting\n");
 	while (1) {
-		scanf(E_F, &input);
-		fifo_put_element(fifo_a_b, 2*input);
-		scanf(E_F, &input);
-		fifo_put_element(fifo_a_c, 3*input);
+		fifo_get_multiple(fifo_in_a, 2, input);
+		fifo_put_element(fifo_a_b, 2*input[0]);
+		fifo_put_element(fifo_a_c, 3*input[1]);
 	}
 	pthread_exit(NULL);
 }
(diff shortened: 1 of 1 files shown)


SamuelMiksits (commented):
Great! The only thing that I wonder about is, why does input need to be its own actor since it being in A already fulfills that all inputs must come from one actor?

klarasm:
> Great! The only thing that I wonder about is, why does input need to be its own actor since it being in A already fulfills that all inputs must come from one actor?

Not needed in this case, no. But I thought that this should show the general principle, meaning that if we want another actor, the method can still be reused.

SamuelMiksits:
I will indulge in some nitpicking:

If I run the program and press Ctrl+D in the current state, it returns:

```
sam@tumbleweed:~/Documents/IL2232/forsyde-devtools/examples/implementation> ./main
Process Input starting
Process A starting
Process B starting
Process C starting
Process D starting
Process Output starting
1 1 1 1 1
out1: 0, out2: 4
out1: 0, out2: 4
Input exiting

sam@tumbleweed:~/Documents/IL2232/forsyde-devtools/examples/implementation> echo $?
130
```
and there is no `return` statement at the end of an `int main`. Should this actually be the case? I feel like Ctrl+D is the natural way to terminate a program in this form and that if it stops successfully it should stop the threading and return exit success. Even if I add `return 0;` at the bottom it still has this `130` return code.

Maybe this is over-engineering but I feel in general like we should only ever return non-zero exit codes if our program got into some undefined state where we cannot guarantee that the output is correct.

klarasm:
> I will indulge in some nitpicking:
> 
> If I run the program and press Ctrl+D in the current state, it returns:
> 
> ```
> sam@tumbleweed:~/Documents/IL2232/forsyde-devtools/examples/implementation> ./main
> Process Input starting
> Process A starting
> Process B starting
> Process C starting
> Process D starting
> Process Output starting
> 1 1 1 1 1
> out1: 0, out2: 4
> out1: 0, out2: 4
> Input exiting
> 
> sam@tumbleweed:~/Documents/IL2232/forsyde-devtools/examples/implementation> echo $?
> 130
> ```
> 
> and there is no `return` statement at the end of an `int main`. Should this actually be the case? I feel like Ctrl+D is the natural way to terminate a program in this form and that if it stops successfully it should stop the threading and return exit success. Even if I add `return 0;` at the bottom it still has this `130` return code.
> 
> Maybe this is over-engineering but I feel in general like we should only ever return non-zero exit codes if our program got into some undefined state where we cannot guarantee that the output is correct.

You're right, I'll add it when I get home.

klarasm:
> I will indulge in some nitpicking:
> 
> If I run the program and press Ctrl+D in the current state, it returns:
> 
> ```
> sam@tumbleweed:~/Documents/IL2232/forsyde-devtools/examples/implementation> ./main
> Process Input starting
> Process A starting
> Process B starting
> Process C starting
> Process D starting
> Process Output starting
> 1 1 1 1 1
> out1: 0, out2: 4
> out1: 0, out2: 4
> Input exiting
> 
> sam@tumbleweed:~/Documents/IL2232/forsyde-devtools/examples/implementation> echo $?
> 130
> ```
> 
> and there is no `return` statement at the end of an `int main`. Should this actually be the case? I feel like Ctrl+D is the natural way to terminate a program in this form and that if it stops successfully it should stop the threading and return exit success. Even if I add `return 0;` at the bottom it still has this `130` return code.
> 
> Maybe this is over-engineering but I feel in general like we should only ever return non-zero exit codes if our program got into some undefined state where we cannot guarantee that the output is correct.

The return value seems to be a result of `pthread_kill`. Switched that to `pthread_cancel` and the exit value stays zero with the return.

SamuelMiksits:
Gotcha, yes I was running your first force push and was still getting 130 exit code. If I put a print statement after the `pthread_kill` for-loop it will not run so I overall agree with that assessment. I can try to rerun.

klarasm:
(I'll just clean up a little. I left an unused ret in some of the processes)

SamuelMiksits:
Okay I think this is pretty good now, ready for merge?

klarasm:
> Okay I think this is pretty good now, ready for merge?

I have no objections

PR #163, opened 2025-10-28
Update IRConstructors to include local inputs and outputs

### LINKED ISSUE #157: ForSyDe IR is Missing Details
new chaos

### What to change
Add a local inputs and outputs ordered list to ForSyDe IR process constructors.

### How to change
A possible solution could be to add a tuple of two string lists to the `IRConstructor` data type, one for local input signal IDs and the other for local output signal IDs:
```haskell
data IRConstructor
  = IRDelay String [Int] ([String], [String])
  | IRActor String ActorType String ([String], [String])
```

### Why to change
There needs to be some connection between how signals relate to the inputs and outputs of a process constructor. This is especially important for actors with multiple inputs and outputs. How would we know if a signal is the first or second output of an actor? Our current ForSyDe IR does not provide an answer.

### When to change
Change can be done before or while implementing parts of the Core IR to ForSyDe IR transformation.

sthaeron:
- Updates ForSyDe IR so process constructors have local inputs and outputs
- Refactors ForSyDe IR testing
- Refactors SDF Scheduling
- Refactors CoreToForSyDeIR
- Cleans up naming and shadowing warning in CoreToForSyDeIR
- Adds CoreToForSyDeIR to cabal

This pull request resolves: #157

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,9 +22,15 @@ cabal.project.local
 cabal.project.local~
 .HTF/
 .ghc.environment.*
+
+# ForSyDe DevTools Specific
 *.hcr
-*.irf
 *.fir
+*.pir
+!/examples/test/*.hcr
+!/examples/test/*.fir
+!/examples/test/*.pir
+
 # Ignore build outputs from performing a nix-build or `nix build` command
 result
 result-*
--- a/examples/test/simple.fir
+++ b/examples/test/simple.fir
@@ -1,8 +1,8 @@
 IRSystem(
   {"input"}, {"output"},
   {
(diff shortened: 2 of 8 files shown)


sthaeron:
@Himawari722 I updated the IRSystem examples in your SDF Scheduling module based on the order of the IRSignals, perhaps you could double check them to see if it makes sense.

Himawari722 (approved):
I had a look at the updated IRSystem examples in scheduling, and they seem correct to me!

MichelD-Abo (approved):
Good!

PR #165, opened 2025-10-28
Work/procedural ir

Himawari722:
I have updated the Procedural IR to support more C language features used in the C code examples and add a corresponding documentation with it.

Some sample extended features are:

- `SFor`: Since for loops are desugared into while loops in Cigrid, I added SFor so that we can avoid using while loops all the time.
- `SNew`, `SDelete` and `SArrayDef`: Since `new` and `delete` are c++ keywords and they are the only way to declare arrays in Cigrid, they are deleted and `SArrayDef` are added to allow `int output[2]`.
- `++` and `--`: Increment and Decrement are added to unary operators. However, current definition in unary operation does not contain information about its associativity, they should be placed in the left the same as all other unary operators: `-x`, `!y`, `++i`. This will be corrent when writing for loops, but use with care in array access

Other added language features are specified in the documentation.


docs/procedural-ir.md
sthaeron, line 1: Procedural IR is not really based on the Cigrid AST but rather the Cigrid IR sometimes refered to as Hybrid IR in the lecture slides. The Hybrid IR just reuses aspects of the Cigrid AST, but also introduces a lot more representations. Should also mention that Procedural IR can represent control-flow graphs.

sthaeron (approved):
Tests worked and the code looks good. I gave the Procedural IR docs a read and it seems fine to me. I had one issue with the doc, which was the comparison to Cigrid AST and suggested changing it to Cigrid IR or Hybrid IR.

docs/procedural-ir.md
Himawari722: Thanks for the feedback! Regarding the Hybrid IR, I have checked the lecture slides and Hybrid IR are suppose to be represented as CFGs, which code are separated in blocks and control flows are transformed into branches between blocks. However the Procedural IR here stays at the syntatic level and it is closer to a tree structure like the AST, that's why I am comparing Procedural IR to Cigrid AST since it utilize the syntaxes to represent and reconstruct the C grammar.

sthaeron: What is our Procedural IR missing from Cigrids Hybrid IR which represents control-flow graphs? If Procedural IR can't represent control-flow graphs and is only focused on reconstructing the C grammar, we might need to have an additional IR before Procedural IR to better support multi-core.

Himawari722: I dont think we need the control-flow graphs to support multi-core? I think control-flow graph are for low-level serial codes, and multi-core is more a task level thing, which is supported by C codes using features like pthreads and mutex. So what am I thinking is that there might be a separate multi-core scheduler that generate multi-core schedules mapping tasks to cores and synchronizations from the ForSyDe IR, and the multi-core ForSyDe IR to Procedural IR module will then generate the corresponding C code. So explicit control-flow graph representation isn’t required at this level. But additional information is indeed required, for example the execution time. We can have a look at the book Ingo recommended to determine what information would be needed to be extendable to multi-core implementations.

klarasm: I agree we probably don't need CFG to generate multi-core code, especially as we are probably going to only do external assignment and primarily do communication through the templates/libraries. However, having CFGs would be nice for other reasons, e.g. if we manage to get to more advanced optimizations. Do we really need it at this stage, though?

sthaeron: I've been convinced and agree we don't currently need an IR that can represent control-flow graphs. I'll merge the pr then and give other pr's the chance to fix any potential conflicts with `forsyde-devtools.cabal`.

PR #166, opened 2025-10-28
docs: add LSP information to visualization.md

klarasm:
Try to put down what we have on LSP so far.

### DIFF
--- a/docs/visualisation.md
+++ b/docs/visualisation.md
@@ -1,18 +1,48 @@
 # Visualisation Documentation
 
+## Overview
+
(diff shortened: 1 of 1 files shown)


sthaeron (approved):
Gave it a read and it sounds good to me!

SamuelMiksits:
How do we deal with this now? Can probably just change some of the "plyghd" stuff to "LSP" and "Python" to "Haskell" and keep it since it works kinda the same.

Just looking through the PRs since we had like 8 open earlier today and trying to get the number down

klarasm:
> How do we deal with this now? Can probably just change some of the "plyghd" stuff to "LSP" and "Python" to "Haskell" and keep it since it works kinda the same.
> 
> Just looking through the PRs since we had like 8 open earlier today and trying to get the number down

Yes, that would be an option. I can start working on that in the next few days probably.

ramadhanafif (approved):
LGTM, I would like to add a bit more on VS Code side and possible code actions.

PR #169, opened 2025-10-29
Compiler binary works independently from dev env

sthaeron:
The binary built from the command `nix build` or run with `nix run` can now be run from outside our dev env without resulting in an error mentioning ForSyDe Shallow is missing!
- Updated nix overlay to extend ghc 9.10.2 haskell packages
- Updated nix overlay to wrap haskell packages into binary
- Added explicit app in nix flake to run

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -5,17 +5,34 @@
   inputs.flake-utils.url = "github:numtide/flake-utils";
 
   outputs =
-    { flake-utils, ... }@inputs:
+    { self, flake-utils, ... }@inputs:
     let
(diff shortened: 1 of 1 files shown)


sthaeron:
Can test it out using the following command, make sure to be in the root of our repo but outside of our development environment (don't run `nix develop`):
```sh
nix run github:sthaeron/forsyde-devtools/fix/flakeBuild -- examples/model/SDF_example_003.hs --output-forsyde-ir --stdout
```

klarasm (approved):
Nice to have this done!

PR #170, opened 2025-10-30
Add a prototype LSP

klarasm:
Prototype LSP. Mostly similar to the current plyghd prototype except some special cases. It now also gets the source file name from the LSP client.

### DIFF
--- a/src/ArgumentsLSP.hs
+++ b/src/ArgumentsLSP.hs
@@ -1,0 +1,73 @@
+module ArgumentsLSP where
+
+import Options.Applicative
+
+-- Which file to use for input
+data Input
+  = InputFile FilePath
+  | FromClient
+
+-- Host IP(V4) address
+data IP
+  = Host String
+
+-- Port for the server
+data Port
+  = TCP String
+
+-- Arguments, kind of horrible names but used to not shadow usage of ip/port
+-- in main. Contains an ip address, port and file name for synthesis in the
+-- server
+data Arguments = Arguments
+  { -- Files
+    hostIp :: IP,
+    serverPort :: Port,
+    -- Formats
+    input :: Input
+  }
+
+-- Handle file input. Default is to get the file from your client, but
+-- allow user to pass a filename directly to server for debugging.
+inputFile :: Parser Input
+inputFile =
+  argument
+    (InputFile <$> str)
+    ( metavar "INPUT"
+        <> value FromClient
+        <> help "Input filename (debug option), if unset get file from client"
+    )
+
+-- IP - String "127.0.0.1" by default, otherwise provide -a or --address and
+-- the ip address
+ipAddress :: Parser IP
+ipAddress =
+  option
+    (Host <$> str)
+    ( short 'a'
+        <> long "address"
+        <> metavar "IP"
+        <> value (Host "127.0.0.1")
+        <> help "Host IP"
+    )
+
+-- Port - String "5007" by default, otherwise provide -p or --port and
+-- the port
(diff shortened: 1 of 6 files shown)


ramadhanafif (approved):

SamuelMiksits:
Can you give an example on how to run this (klighd-linux cli and forsyde-lsp commands) and what you expect it to do? I can't figure it out and just get some JSON rpc error but I am not sure how I am supposed to run it

klarasm:
I think it should be sufficient with:
```sh
cabal run forsyde-lsp-exe
# in another terminal
./klighd-cli --ls_port 5007 <path to a forsyde model>
```
If the browser is not opened you can go to the thing which klighd-cli outputs

src/ArgumentsLSP.hs
```diff
@@ -0,0 +1,79 @@
...
+-- in main. Contains an ip address, port and file name for synthesis in the
+-- server
+data Arguments = Arguments
+  { -- Files
+    hostIp :: IP,
+    serverPort :: Port,
+    -- Formats
+    input :: Input
+  }
+
```
SamuelMiksits, line 28: Ah my bad. It took me like an hour of debugging to implement `inputFile` where I tried different approaches, among them calling `fileName` from `inputFile`. But it ended up being the case that I didn't need this function. So can literally just

```diff
diff --git a/src/ArgumentsLSP.hs b/src/ArgumentsLSP.hs
index 75a2967..72f6ef4 100644
--- a/src/ArgumentsLSP.hs
+++ b/src/ArgumentsLSP.hs
@@ -26,12 +26,6 @@ data Arguments = Arguments
     input :: Input
   }
 
--- Function to create a filepath with file extension
-fileName :: ReadM (FilePath)
-fileName = str >>= returnFile
-  where
-    returnFile s = pure s
-
```

SamuelMiksits (approved):
Solid prototype, has some issues with rendering that I think we handled in Python that are not handled here yet, could be addressed with `case` in the future:
- Edge port should not have a token rate when:
  - Source is global input
  - Destination is global output
  - Source is delay
  - Destionation is delay
- Edge port should not have a "arrow" when destination is delay
- Size constraint is weird for delay, so usually ends up being an ellipsoid rather than a circle. But we were considering if we should instead add delay as an "ascii ball" edge label with inline label setting which would fix the other delay issues that I mentioned. But could also interfere with our edge labels (like "s_1" etc)? I forget how it is implemented, you might be able to just put the inline label constraint on that label on top of having the "s_1" label that acts normally. 

sdf_002

klarasm:
> Solid prototype, has some issues with rendering that I think we handled in Python that are not handled here yet, could be addressed with `case` in the future:
> 
>    * Edge port should not have a token rate when:
>      
>      * Source is global input
>      * Destination is global output
>      * Source is delay
>      * Destionation is delay
>
>    * Edge port should not have a "arrow" when destination is delay
>
>    * Size constraint is weird for delay, so usually ends up being an ellipsoid rather than a circle. But we were considering if we should instead add delay as an "ascii ball" edge label with inline label setting which would fix the other delay issues that I mentioned. But could also interfere with our edge labels (like "s_1" etc)? I forget how it is implemented, you might be able to just put the inline label constraint on that label on top of having the "s_1" label that acts normally.

@SamuelMiksits:

I think I resolved most of these now. I kept the delay as an ellipse but removed some size constraints, made it smaller, added background color, and removed the label. I also added the signal labels.

SDF_example_002(1)

klarasm:
I'll merge this now so other people can work on it more easily.

PR #171, opened 2025-10-30
Some general improvement to the CI

klarasm:
Allow to skip CI.

Only build/test when relevant files change.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Build and test the compiler
         run: |
           git fetch origin dev
-          if git diff --stat origin/dev | grep '\(^ src\|^ test\|^ forsyde-devtools.cabal\|^ cabal.project\)'; then
+          if git diff --stat origin/dev | grep '\(^ src\|^ test\|^ flake.nix\|^ flake.lock\|^ forsyde-devtools.cabal\|^ cabal.project\)'; then
             nix develop --command bash -c "cabal build"
             nix develop --command bash -c "cabal test"
           fi
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Build and test the compiler
         run: |
           git fetch origin dev
-          if git diff --stat origin/dev | grep '\(^ src\|^ test\|^ flake.nix\|^ flake.lock\|^ forsyde-devtools.cabal\|^ cabal.project\)'; then
+          if git diff --stat origin/dev -- | grep '\(^ src\|^ test\|^ flake.nix\|^ flake.lock\|^ forsyde-devtools.cabal\|^ cabal.project\)'; then
             nix develop --command bash -c "cabal build"
             nix develop --command bash -c "cabal test"
           fi
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,6 +19,8 @@ jobs:
   build:
     # The type of runner that the job will run on
     runs-on: ubuntu-latest
+    # Allow skipping ci
+    if: "!contains(github.event.pull_request.title, '[ci skip]') && !contains(github.event.pull_request.body, '[ci skip]')"
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,5 +32,8 @@ jobs:
         # Runs a set of commands using the runners shell
       - name: Build and test the compiler
         run: |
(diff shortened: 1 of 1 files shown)


ramadhanafif:
Looks good, maybe I can try it in my fork first

sthaeron:
I added `flake.nix` and `flake.lock` to the list of build if changed since if our dev enviroment or dependency versions change we should probably test and rebuild on the CI as well. Nevertheless, when pushing my change to this pr I checked the build logs of the CI and it seems there is a shell error:
```
fatal: ambiguous argument 'origin/dev..HEAD': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```
I tested the shell if statement as a separate shell script and it worked on my computer. Seems to be a CI specific problem. Note sure were they want us to add `--` either.

klarasm:
> I added `flake.nix` and `flake.lock` to the list of build if changed since if our dev enviroment or dependency versions change we should probably test and rebuild on the CI as well. Nevertheless, when pushing my change to this pr I checked the build logs of the CI and it seems there is a shell error:
> 
> ```
> fatal: ambiguous argument 'origin/dev..HEAD': unknown revision or path not in the working tree.
> Use '--' to separate paths from revisions, like this:
> 'git <command> [<revision>...] -- [<file>...]'
> ```
> 
> I tested the shell if statement as a separate shell script and it worked on my computer. Seems to be a CI specific problem. Note sure were they want us to add `--` either.

I think the remote might be named something else or similar. The `--` is only if we want to limit to certain files I believe.

klarasm:
Could you try if it works now? (changed from origin/dev to dev).

sthaeron:
You can check the github actions logs it ran the CI and it failed the if statement again with the same error but now its `dev`:
```
fatal: ambiguous argument 'dev..HEAD': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

sthaeron:
I tested adding `--` to the end of `git diff --stat dev..HEAD` and before the grep. Ended up getting a shorter error than before:
```
fatal: bad revision 'dev..HEAD'
```

klarasm:
> I tested adding `--` to the end of `git diff --stat dev..HEAD` and before the grep. Ended up getting a shorter error than before:
> 
> ```
> fatal: bad revision 'dev..HEAD'
> ```

It might be that the checkout hook does not fetch the dev branch. I added that now.

klarasm:
This seems to be working now:
```
Run git fetch origin dev
  git fetch origin dev
  if git diff --stat origin/dev -- | grep '\(^ src\|^ test\|^ flake.nix\|^ flake.lock\|^ forsyde-devtools.cabal\|^ cabal.project\)'; then
    nix develop --command bash -c "cabal build"
    nix develop --command bash -c "cabal test"
  fi
  shell: /usr/bin/bash -e {0}
  env:
    NIX_PATH: nixpkgs=channel:25.05
    TMPDIR: /home/runner/work/_temp
    NIX_PROFILES: /nix/var/nix/profiles/default /home/runner/.nix-profile
    NIX_SSL_CERT_FILE: /etc/ssl/certs/ca-certificates.crt
From https://github.com/sthaeron/forsyde-devtools
 * branch            dev        -> FETCH_HEAD
 * [new branch]      dev        -> origin/dev
 ```

**Comment by klarasm** [comment]
@sthaeron I think right now it only checks the description of the PR, not the commits.

**Comment by sthaeron** [comment]
> @sthaeron I think right now it only checks the description of the PR, not the commits.

Oh really, well it did work for the commit as well it did not run the ci when I added `[ci skip]`. Reread the hook you updated and you're write to only relates to the pr. I'll have to double check it again then. I plan on dropping these test commits in a rebase later.

**Comment by sthaeron** [comment]
See my last commit had a `[ci skip]` and it did not start a CI build. Maybe commits messages are part of the pr body since you are checking both: `if: "!contains(github.event.pull_request.title, '[ci skip]') && !contains(github.event.pull_request.body, '[ci skip]')"`.

**Comment by sthaeron** [comment]
Found why it works, its part of githubs default functionality at the following link https://docs.github.com/en/actions/how-tos/manage-workflow-runs/skip-workflow-runs it mentions:
```text
Workflows that would otherwise be triggered using on: push or on: 
pull_request won't be triggered if you add any of the following strings to the commit message in a push, 
or the HEAD commit of a pull request:

[skip ci]
[ci skip]
[no ci]
[skip actions]
[actions skip]
```

sthaeron (approved):
After our changes and testing the feature, everything seems to be working as expected. I rebased the pr to delete my testing commits. I feel like this pr is ready to be merged.

PR #172, opened 2025-11-02
Documentation for C code generation with actor templates and libraries

SamuelMiksits:


SamuelMiksits:
(Adding Klara and Zicong for review since Klara also looked at the platform independent C example and Zicong is also working on defining procedural IR)

Himawari722:
It looks good and detailed! The only thing I am not sure about is:
```
#if PLATFORM == PICO2
        // Initialize all components on the lab-kit.
        BSP_Init();
#endif
```
Do we want to keep this code in every C code we generated, regardless of the target platform? Does the arguments takes target platform now, so that codegen part can know whether it should add `BSP_Init()`?

klarasm:
> It looks good and detailed! The only thing I am not sure about is:
> 
> ```
> #if PLATFORM == PICO2
>         // Initialize all components on the lab-kit.
>         BSP_Init();
> #endif
> ```
> 
> Do we want to keep this code in every C code we generated, regardless of the target platform? Does the arguments takes target platform now, so that codegen part can know whether it should add `BSP_Init()`?

I haven't read thoroughly yet, but for this part we could keep the ifdeffery in the headers and just have a generic `init()` function. Then we would have something like this in `common.h` instead:
```c
#if PLATFORM == PC
static void init(void) {}
#elif PLATFORM == PICO2
static void init(void) {
    BSP_Init();
}
#endif
```

SamuelMiksits:
> It looks good and detailed! The only thing I am not sure about is:
> 
> ```
> #if PLATFORM == PICO2
>         // Initialize all components on the lab-kit.
>         BSP_Init();
> #endif
> ```
> 
> Do we want to keep this code in every C code we generated, regardless of the target platform? Does the arguments takes target platform now, so that codegen part can know whether it should add `BSP_Init()`?

Ah, that is a good question actually. I see your point, which is that you can argument parse and use that to figure out whether the code should be that in the first place or not.

In the example `example/implementations/platform_independent/SDF_example_002/` I have the file `common.h` which contains:

```C
/** @brief Selected target platform is a Linux PC */
#define PC 1

/** @brief Selected target platform is a Pico 2 board */
#define PICO2 2

/**  @brief Platform modes, use this parameter to select platform.
 *   Available platforms so far:
 *  @param PC - Linux PC (pthreads)
 *  @param PICO2 - PICO 2 board
 */
#ifndef PLATFORM
#define PLATFORM PC
#endif

#if PLATFORM == PC
    #include "buffer_blocking_pc.h"
#endif

#if PLATFORM == PICO2
    #include "buffer_blocking_pico2.h"
    #include <bsp.h>
#endif
```

So the idea is that currently, argument parsing does not take target platform into account (can be added), so after running our compiler, you compile the output C with

```
> gcc ./main.c -DPLATFORM=PC
```

for PC and

```
> gcc ./main.c -DPLATFORM=PICO2
```

for Pico 2 board. (in actuality you add 

```cmake
add_executable(LabKitTest main_platform_independent.c ${BSP_SOURCES})
target_compile_options(LabKitTest PRIVATE -DPLATFORM=PICO2)
```
to your CMake in the lab kit)

and these issues get resolved at "compile time" (perhaps poor phrasing since we are building a compiler haha, "gcc compile time").

I tried compiling for "PC" on the Pico 2 board, and after just removing the include for pthreads blocking buffers I could compile it. Then I tried running on the Pico 2 board without the `BSP_Init()` line and when I opened the serial monitor and started to send messages, nothing would happen. So scanf/printf was not working without it.

Himawari722:
I think Klara’s idea is pretty neat, since right now I have implemented Procedural IR to C which is pure grammar parts and platform unrelated, so it would be nice to separate the core logics(in main function) and platform related things

SamuelMiksits:
> I haven't read thoroughly yet, but for this part we could keep the ifdeffery in the headers and just have a generic `init()` function. Then we would have something like this in `common.h` instead:
> 
> ```c
> #if PLATFORM == PC
> static void init(void) {}
> #elif PLATFORM == PICO2
> static void init(void) {
>     BSP_Init();
> }
> #endif
> ```

Yeah this is a good idea, I think maybe something like:

```C
static void init()
{
  #if PLATFORM == PICO2
    BSP_Init();
  #endif
}
```

and keep adding them there might be "cleaner". But you could also take an argument into the compiler and make the code gen add the lines

```C
#define PLATFORM PC
```
or
```C
#define PLATFORM PICO2
```
in `main.c` either just in general (while keeping this ifdef-stuff) so you don't need to bother with the `-DPLATFORM` compile options, or to handle the includes and then make the code gen either not add the `BSP_Init()` at all for PC or add it on Pico 2 (and then you however you might need to do something for future platforms).

SamuelMiksits:
Okay I did another force push where I added
```C
init()
```
and then changed `common.h` so that we have
```C
#if PLATFORM == PC
    #include "buffer_blocking_pc.h"
    static void init() {}
#endif

#if PLATFORM == PICO2
    #include "buffer_blocking_pico2.h"
    #include <bsp.h>
    static void init() {
        BSP_Init();
    }
#endif
```
I implemented it this way instead of making one `init()` function where I put separate `#if` inside again

klarasm:
> Okay I did another force push where I added
> 
> ```c
> init()
> ```
> 
> and then changed `common.h` so that we have
> 
> ```c
> #if PLATFORM == PC
>     #include "buffer_blocking_pc.h"
>     static void init() {}
> #endif
> 
> #if PLATFORM == PICO2
>     #include "buffer_blocking_pico2.h"
>     #include <bsp.h>
>     static void init() {
>         BSP_Init();
>     }
> #endif
> ```
> 
> I implemented it this way instead of making one `init()` function where I put separate `#if` inside again

I'm a bit nitpicky again, but I would prefer to have `static void init(void)`. In C standards before C23 the version with `static void init()` without the `void` can accept any number of parameters, e.g:
```C
int test() {
    return 0;
}

int main(int argc, char **argv) {
    return test(1, 2, 3);
}
```
and similar are valid. The pico toolchain currently uses GCC 14 which does not yet have C23 as the default (I believe this was changed in GCC 15).

SamuelMiksits:
> In C standards before C23 the version with `static void init()` without the `void` can accept any number of parameters, e.g:
> 
> ```c
> int test() {
>     return 0;
> }
> 
> int main(int argc, char **argv) {
>     return test(1, 2, 3);
> }
> ```
> 
> and similar are valid. The pico toolchain currently uses GCC 14 which does not yet have C23 as the default (I believe this was changed in GCC 15).

Really? Okay that is pretty stupid imo, I can add `void`

klarasm:
> > In C standards before C23 the version with `static void init()` without the `void` can accept any number of parameters, e.g:
> > ```c
> > int test() {
> >     return 0;
> > }
> > 
> > int main(int argc, char **argv) {
> >     return test(1, 2, 3);
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and similar are valid. The pico toolchain currently uses GCC 14 which does not yet have C23 as the default (I believe this was changed in GCC 15).
> 
> Really? Okay that is pretty stupid imo, I can add `void`

Yep it's pretty horrible, things like this is also allowed before C23 (edit: hmm, this seems to still "work" in C23) and compiles but introduces UB:
```C
int test(a, b, c)
	int a, b, c;
{
	return c;
}

int main(int argc, char **argv)
{
	return test(1, 2);
}
```

SamuelMiksits:
Okay now we have

```C
#if PLATFORM == PC
    #include "buffer_blocking_pc.h"
    static void init(void) {}
#endif

#if PLATFORM == PICO2
    #include "buffer_blocking_pico2.h"
    #include <bsp.h>
    static void init(void) {
        BSP_Init();
    }
#endif
```

I can convert to non-draft PR and if you are happy either of you can merge

klarasm (approved):
I looked through this quickly again and it seems good!

Himawari722:
LGTM!

Himawari722 (approved):

PR #173, opened 2025-11-03
Feat/core to forsy de ir

sthaeron:
Implements Core IR to ForSyDe IR for the following programs:
- Actor22 and a Delay, using a simple add function
- Actor12, using a simple add function


MichelD-Abo (approved):
Good!

PR #177, opened 2025-11-04
Adding basic actors

sthaeron:
Added SDF Actor11 and Actor21. Working `SDF_example_002.hs`. Pretty hacky and will require further changes for more complex net lists.

### DIFF
--- a/examples/model/SDF_example_001.hs
+++ b/examples/model/SDF_example_001.hs
@@ -1,31 +1,26 @@
--- vi: ts=4 sts=4 sw=4 et
 module SDF_example_001 where
 
 import ForSyDe.Shallow
 
-system :: Signal Int -> (Signal Int, Signal Int)
-system s_in = (s_out1, s_out2) where
-    -- Process A
-    (s_a_b, s_a_c) = processA s_in
-    -- Process B
-    (s_out1, s_b_d) = processB s_a_b
-    -- Process C
-    s_c_d = processC s_a_c
-    -- Process D
-    s_out2 = processD s_b_d s_c_d
+system s_in = (s_out1, s_out2)
+  where
+    (s_1_2, s_1_3) = a_1 s_in
+    (s_out1, s_2_4) = a_2 s_1_2
+    s_3_4 = a_3 s_1_3
+    s_out2 = a_4 s_2_4 s_3_4
 
-processA :: Signal Int -> (Signal Int, Signal Int)
-processA s_in = actor12SDF 2 (1,1) f s_in where
-    f [a, b] = ([2*a], [3*b])
+a_1 s_in = actor12SDF 2 (1, 1) f_1 s_in
 
-processB :: Signal Int -> (Signal Int, Signal Int)
-processB s_in = actor12SDF 1 (1, 1) f s_in where
-    f [a] = ([a-2], [a-1])
+a_2 s_in = actor12SDF 1 (1, 1) f_2 s_in
 
-processC :: Signal Int -> Signal Int
-processC s_in = actor11SDF 1 1 f s_in where
-    f [a] = [a+1]
+a_3 s_in = actor11SDF 1 1 f_3 s_in
 
-processD :: Signal Int -> Signal Int -> Signal Int
-processD s_in1 s_in2 = actor21SDF (1, 1) 1 f s_in1 s_in2 where
-    f [a] [b] = [a*b]
+a_4 s_in1 s_in2 = actor21SDF (1, 1) 1 f_4 s_in1 s_in2
+
+f_1 [x, y] = ([2 * x], [3 * y])
+
+f_2 [x] = ([x - 2], [x - 1])
+
+f_3 [x] = [x + 1]
+
+f_4 [x] [y] = [x * y]
--- a/examples/model/SDF_example_001.hs
+++ b/examples/model/SDF_example_001.hs
@@ -2,21 +2,24 @@ module SDF_example_001 where
 
 import ForSyDe.Shallow
 
-system s_in = (s_out1, s_out2)
+-- Net list
+system s_in = (s_out_1, s_out_2)
   where
-    (s_1_2, s_1_3) = a_1 s_in
-    (s_out1, s_2_4) = a_2 s_1_2
-    s_3_4 = a_3 s_1_3
-    s_out2 = a_4 s_2_4 s_3_4
+    (s_1_2, s_1_3) = a_a s_in
+    (s_out_1, s_2_4) = a_b s_1_2
+    s_3_4 = a_c s_1_3
+    s_out_2 = a_d s_2_4 s_3_4
 
-a_1 s_in = actor12SDF 2 (1, 1) f_1 s_in
+-- Process specifications
+a_a s_in = actor12SDF 2 (1, 1) f_1 s_in
 
-a_2 s_in = actor12SDF 1 (1, 1) f_2 s_in
+a_b s_in = actor12SDF 1 (1, 1) f_2 s_in
 
-a_3 s_in = actor11SDF 1 1 f_3 s_in
+a_c s_in = actor11SDF 1 1 f_3 s_in
 
-a_4 s_in1 s_in2 = actor21SDF (1, 1) 1 f_4 s_in1 s_in2
+a_d s_in_1 s_in_2 = actor21SDF (1, 1) 1 f_4 s_in_1 s_in_2
 
+-- Function definitions
 f_1 [x, y] = ([2 * x], [3 * y])
 
 f_2 [x] = ([x - 2], [x - 1])
--- a/examples/model/SDF_example_002.hs
+++ b/examples/model/SDF_example_002.hs
@@ -3,32 +3,30 @@ module SDF_example_002 where
 import ForSyDe.Shallow
 
 -- Netlist
-system :: Signal Int -> Signal Int -> Signal Int
-system s_ina s_inb = s_out where
-    s_1 = actor_a s_ina
-    s_2 = actor_b s_inb
-    s_3 = actor_c s_1 s_4_delayed
-    (s_4, s_out) = actor_d s_2 s_3
+system s_ina s_inb = s_out
+  where
+    s_1 = a_1 s_ina
+    s_2 = a_2 s_inb
+    s_3 = a_3 s_1 s_4_delayed
+    (s_out, s_4) = a_4 s_2 s_3
     s_4_delayed = d_1 s_4
 
 -- Process specifications
-actor_a :: Signal Int -> Signal Int
-actor_a s1 = actor11SDF 2 1 f_1 s1
-actor_b :: Signal Int -> Signal Int
-actor_b s1 = actor11SDF 1 2 f_2 s1
-actor_c :: Signal Int -> Signal Int -> Signal Int
-actor_c s1 s2 = actor21SDF (2,1) 1 f_3 s1 s2
-actor_d :: Signal Int -> Signal Int -> (Signal Int, Signal Int)
-actor_d s1 s2 = actor22SDF (2,1) (1,2) f_4 s1 s2
-d_1 :: Signal Int -> Signal Int
-d_1 s = delaySDF [0] s 
+a_1 s1 = actor11SDF 2 1 f_1 s1
+
+a_2 s1 = actor11SDF 1 2 f_2 s1
+
+a_3 s1 s2 = actor21SDF (2, 1) 1 f_3 s1 s2
+
+a_4 s1 s2 = actor22SDF (2, 1) (1, 2) f_4 s1 s2
+
+d_1 s = delaySDF [0] s
 
 -- Function definitions
-f_1 :: [Int] -> [Int]
 f_1 [x, y] = [x + y]
-f_2 :: [Int] -> [Int]
+
 f_2 [x] = [x, x + 1]
-f_3 :: [Int] -> [Int] -> [Int]
+
 f_3 [x, y] [z] = [x + y + z]
-f_4 :: [Int] -> [Int] -> ([Int], [Int])
+
 f_4 [x, y] [z] = ([x + y + z], [x + y, x + y + z])
--- a/examples/model/SDF_example_002.hs
+++ b/examples/model/SDF_example_002.hs
@@ -3,22 +3,22 @@ module SDF_example_002 where
 import ForSyDe.Shallow
 
 -- Netlist
-system s_ina s_inb = s_out
+system s_in_1 s_in_2 = s_out
   where
-    s_1 = a_1 s_ina
-    s_2 = a_2 s_inb
-    s_3 = a_3 s_1 s_4_delayed
-    (s_out, s_4) = a_4 s_2 s_3
+    s_1 = a_a s_in_1
+    s_2 = a_b s_in_2
+    s_3 = a_c s_1 s_4_delayed
+    (s_out, s_4) = a_d s_2 s_3
     s_4_delayed = d_1 s_4
 
 -- Process specifications
-a_1 s1 = actor11SDF 2 1 f_1 s1
+a_a s_1 = actor11SDF 2 1 f_1 s_1
 
-a_2 s1 = actor11SDF 1 2 f_2 s1
+a_b s_1 = actor11SDF 1 2 f_2 s_1
 
-a_3 s1 s2 = actor21SDF (2, 1) 1 f_3 s1 s2
+a_c s_1 s_2 = actor21SDF (2, 1) 1 f_3 s_1 s_2
 
-a_4 s1 s2 = actor22SDF (2, 1) (1, 2) f_4 s1 s2
+a_d s_1 s_2 = actor22SDF (2, 1) (2, 1) f_4 s_1 s_2
 
 d_1 s = delaySDF [0] s
 
--- a/examples/test/double.hcr
+++ b/examples/test/double.hcr
@@ -1,0 +1,126 @@
+NonRec(add = 
+  Lam(a -> 
+  Lam($dNum -> 
+  Lam(ds -> 
+  Let(
+    NonRec(fail = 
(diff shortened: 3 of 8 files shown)


examples/model/SDF_example_002.hs
```diff
@@ -3,32 +3,30 @@ module SDF_example_002 where
...
 
 -- Netlist
-system :: Signal Int -> Signal Int -> Signal Int
-system s_ina s_inb = s_out where
-    s_1 = actor_a s_ina
-    s_2 = actor_b s_inb
-    s_3 = actor_c s_1 s_4_delayed
-    (s_4, s_out) = actor_d s_2 s_3
+system s_ina s_inb = s_out
+  where
```
SamuelMiksits, line 7: This looks a bit weird in the diff, I guess this is just an ormolu thing?

examples/test/double.hcr
```diff
@@ -0,0 +1,126 @@
...
+  ))))
+)
+
+NonRec($trModule = 
+  App(App(Var(Module) * 
+  App(Var(TrNameS) * 
+  Lit("main"#))) * 
+  App(Var(TrNameS) * 
+  Lit("SDF_example_005"#)))
+)
```
SamuelMiksits, line 126: No newline at end of file. Do we care? I don't want you to add one manually obviously, should we add another newline somewhere in our output, so like instead of having in `Main.hs` (line 44)

```hs
write_output :: Output -> OutputFormat -> [Char] -> IO ()
write_output StdOut _ s =
  putStrLn s
write_output (OutputFile (output_file, WithFileExtension)) _ s =
  writeFile output_file s
write_output (OutputFile (output_file, NoFileExtension)) extension s =
  writeFile (generateDefaultFileName output_file extension) s
```
have:
```hs
write_output :: Output -> OutputFormat -> [Char] -> IO ()
write_output StdOut _ s =
  putStrLn s
write_output (OutputFile (output_file, WithFileExtension)) _ s =
  writeFile output_file (s ++ "\n")
write_output (OutputFile (output_file, NoFileExtension)) extension s =
  writeFile (generateDefaultFileName output_file extension) (s ++ "\n")
```
I tried this and it works, and the extra newline is not really needed for stdout

examples/model/SDF_example_001.hs
```diff
@@ -1,31 +1,26 @@
...
-    f [a] = ([a-2], [a-1])
-
-processC :: Signal Int -> Signal Int
-processC s_in = actor11SDF 1 1 f s_in where
-    f [a] = [a+1]
-
-processD :: Signal Int -> Signal Int -> Signal Int
-processD s_in1 s_in2 = actor21SDF (1, 1) 1 f s_in1 s_in2 where
-    f [a] [b] = [a*b]
+system s_in = (s_out1, s_out2)
```
SamuelMiksits: In general I am not happy with the actor naming convention, but understand you want to keep them consistent between files. I think this convention makes it hard to tell what is an actor and what is a signal in the ForSyDe IR since everything has the exact same "snake shape". In general we should also think about consistency for enumeration, where you name the first signal (internally) with `s_0` but for functions, delays, actors all of  them start with `f_1`, `d_1`, `a_1`

Currently you are proposing:
Signals: `s_x`, x = 0, 1, 2, ... (1, 2, 3, ... in the graph but 0, 1, 2, ... internally)
Actors: `a_x`, x = 1, 2, 3, ...
Delays: `d_x`, x = 1, 2, 3, ...
Functions: `f_x`, x = 1, 2, 3, ...

I think it would be better to have:

Signals: `s_x`, x = 0, 1, 2, ...
Actors: `x`, x = a, b, c, ...
Delays: `d_x`, x = 0, 1, 2, ...
Functions: `f_x`, x = 0, 1, 2, ...

if the intent is to establish some kind of convention.

src/CoreToForSyDeIR.hs
SamuelMiksits, line 1: I think generally this file suffers from lack of comments, and this makes it very hard for me to understand what I am working with if I want to add something.

You should add a comment above the line that contains `data TranslationContext` and explain the following:
- What is the translation context?
- Most parameters are quite obvious, but can you add a comment that explains `nameCounter`?
- Give an example of how to add something to the context
- Give an example of how to query something from the context

The code should contain more comments, but this is just one of the main examples I want to highlight. I can't review things that aren't part of the diff as well, and I don't think I should.

src/CoreToForSyDeIR.hs
```diff
@@ -272,7 +296,9 @@ createDelaySDF context binder expr =
 
 getActorSplit :: ActorType -> Int
 getActorSplit actorType = case actorType of
+  Actor11 -> 1
```
SamuelMiksits: Now it is maybe a bit more obvious, but could you add a comment above `getActorSplit` that says like
```hs
-- Split list of signals between inputs and outputs at this index,
-- should be equal to X in actorXY.
```
or whatever?

src/CoreToForSyDeIR.hs
```diff
@@ -233,32 +234,55 @@ getSourceRate :: TranslationContext -> String -> Int
 getSourceRate context id =
   if elem id (systemInputs context)
     then 1
-    else getRateFromConstructors id (map snd (constructors context))
-
-getRateFromConstructors :: String -> [IRConstructor] -> Int
-getRateFromConstructors id list = case list of
-  [] -> -1
-  prHead : prTail -> case prHead of
```
SamuelMiksits, line 241: What does `pr` mean? (`prHead`, `prTail`, `prId`). I don't understand what this variable is supposed to convey.

examples/model/SDF_example_002.hs
```diff
@@ -3,32 +3,30 @@ module SDF_example_002 where
...
+    s_1 = a_1 s_ina
+    s_2 = a_2 s_inb
+    s_3 = a_3 s_1 s_4_delayed
+    (s_out, s_4) = a_4 s_2 s_3
```
SamuelMiksits: Why did you change the order? this is not correct

sdf_002_tikz

src/CoreToForSyDeIR.hs
```diff
@@ -233,32 +234,55 @@ getSourceRate :: TranslationContext -> String -> Int
...
+      [] -> error ("getSourceRate: " ++ id ++ " not in a valid constructor")
+      prHead : prTail -> case prHead of
+        IRDelay prId _ (_, _) ->
+          if prId == id
+            then 1
+            else aux prTail
+        IRActor prId _ _ (_, _) ->
+          if prId == id
+            then
+              let outputRates = case (findRates prId (actors context)) of
```
SamuelMiksits: I don't really understand this. So basically, does this code only work for `SDF_002.hs` since actor A and actor C both only have 1 output (A being actor11SDF and C actor21SDF)?

I am doing some very hard thinking, but is it actually possible to solve this problem without putting the list of token consumption/production rates in the IRActor in addition to the edge (signal) names?

SamuelMiksits (changes requested):
- I don't like the change in actor naming conventions, but that is not something that "has to be fixed". 

- In general I think the `CoreToForSyDeIR.hs` file needs more comments that explain what the code is doing and how to interact with it. In its current state it can really only be understood by the people who wrote it.

- The order on the "D" actor (see comment) was changed, fix that.

- Could take a look at my proposed solution for no newline at end of file.

- I am not sure what to do about it, but need to think about solutions on how to deal with getting source rate for signals when a signal goes from actor to actor, and the source rate comes from the output of an actor where the actor has multiple outputs.

src/CoreToForSyDeIR.hs
sthaeron: We do have access to the input and ouput rates of actors, its supplied within the `TranslationContext` through the `actors` variable. The problem is that we do not know which of the outputs of an actor actually feed into another actor. Mostly because we haven't started translating complex net lists and reading the core output to find a solution.

examples/model/SDF_example_002.hs
sthaeron: The change is mainly due to how signals are added to IRActors at the end of the translation when building the IRSystem. The current approach will always have global output signals listed as the first output of an actor since the global output signals are translated first.

src/CoreToForSyDeIR.hs
sthaeron: You are absolutely write, if we want others to be able to add to the `CoreToForSyDeIR` translation there should be some comments explaining the core contexts and though process. Do you think every function needs a comment above it?

src/CoreToForSyDeIR.hs
sthaeron: That's my bad, it should actually be `pc` for process constructor. Is it still to vague, should we change it to `constructor` as the prefix instead?

examples/model/SDF_example_001.hs
sthaeron: I only made the change to make it easier for me to read the core output, based on what I was used to. Nevertheless, I am by no means opposed to having an actually established convention for writing our test haskell examples. Your suggestion seems fine to me, but maybe we shouldn't be zero indexing. Once we no longer have to debug core the naming of actors, delays, and functions will be up to the user.

examples/model/SDF_example_002.hs
SamuelMiksits: My issue is that the rate is 2 from D -> out in the graph and 1 from D -> delay. So you have to change the token rate from (1,2) to (2,1) if you have to have it that way, because now the ForSyDe IR has
```
  {
    IRSignal("s_7", ("c", 1), ("d", 1)),
    IRSignal("s_6", ("b", 2), ("d", 2)),
    IRSignal("s_5", ("s_inb", 1), ("b", 1)),
    IRSignal("s_4", ("d_1", 1), ("c", 1)),
    IRSignal("s_3", ("a", 1), ("c", 2)),
    IRSignal("s_2", ("s_ina", 1), ("a", 2)),
    IRSignal("s_1", ("d", 2), ("d_1", 1)),
    IRSignal("s_0", ("d", 1), ("s_out", 1))
  },
```
where D -> delay is 2 and D -> out is 1 which is clearly wrong if you look at D in the graph.
So if you want this, you also need to change from
```diff
- a_4 s1 s2 = actor22SDF (2, 1) (1, 2) f_4 s1 s2
+ a_4 s1 s2 = actor22SDF (2, 1) (2, 1) f_4 s1 s2
```
since you changed the port order

examples/test/double.hcr
sthaeron: Note that the files in the `examples/test` directory are specifically designed to check the pretty printing in test specs. The problem is that when using haskell to read a file as a string it adds a newline at the end. This newline needs to be stripped before comparing the actual pretty print string obtained from our compiler. Note that the test spec bypasses your argument parsing code.

But for the actual compiler using the argument parsing, I'm not sure it really matters if we have a newline or not. Whats the convention in C do you have a newline at the end of your files?

examples/model/SDF_example_002.hs
sthaeron: Ah yes you're right, I'll make the change then

src/CoreToForSyDeIR.hs
SamuelMiksits: No, probably not for all functions. It is generally very hard to know how many comments to add. I just go on feel and don't really think about it, but I think my thought process results in "will I understand this code if I come back to it in 4 years?" and if not due to something very technical way it works then I would put a comment there.

src/CoreToForSyDeIR.hs
SamuelMiksits: Ah okay, can probably change it to maybe `pc` then, I suppose the issue in general is that the function is called `getRateFromConstructors` and not `getRateFromProcessConstructors` and also it is called `IRConstructor` and not `IRProcessConstructor` so "process" seems pulled out of thin air. I can suggest maybe `conHead`, `conTail` and so on maybe, I think using `cons` would be bad since it already means something else in the context of lists.

examples/test/double.hcr
SamuelMiksits: A lot of tools complain if you don't have a newline at the end of the file, but I am not really an expert on why. I also tend to forget to add them, see exhibit `SDF_example_002.hs`. Reading a bit about the topic online, it seems like POSIX defines a line as:

>3.206 Line
>        A sequence of zero or more non-newline characters plus a terminating newline character.

so UNIX-like tools are expecting this to be the case to work correctly. So in our comparison we are using Haskell to compare the lines, but if we built a bash-script test that ran our compiler and compared the output with a tool like `diff` it might not work correctly.

Maybe I am not understanding it correctly, but do you think Haskell is adding it since it is missing in our file? In that case it seems weird to try to work around this by stripping it out and instead rather add it so that both things that are compared should have it already. Another issue could be that in main I am using:

```hs
write_output :: Output -> OutputFormat -> [Char] -> IO ()
write_output StdOut _ s =
  putStrLn s
```
where `putStrLn` automatically adds a newline at the end, perhaps we want to use `putStr` instead.

@klarasm do you have any input on this? Asking since you generally have some opinions on newline at end of file, and you and Sebastian have added tests that drop this newline.

src/CoreToForSyDeIR.hs
SamuelMiksits: Okay, I think we can leave it like this for now, since it is "broken, but less broken than before". Create an issue where you describe this limitation.

examples/test/double.hcr
klarasm: Oh, that's the reason you needed to strip the newline @sthaeron earlier for the tests. Yeah, I don't like it now that I know why. I believe since VSCode originated on windows, it uses the windows convention that newlines are line separators, and thus a final newline will be displayed as an additional line at the end of the file with default settings. I think we should stick with the UNIX/POSIX convention instead.

examples/model/SDF_example_001.hs
```diff
@@ -1,31 +1,29 @@
--- vi: ts=4 sts=4 sw=4 et
```
klarasm, line 1: Oops, I left this one even after I added the .editorconfig, my bad. Good that you remove it.

examples/model/SDF_example_001.hs
```diff
@@ -1,31 +1,29 @@
...
-
-processB :: Signal Int -> (Signal Int, Signal Int)
-processB s_in = actor12SDF 1 (1, 1) f s_in where
-    f [a] = ([a-2], [a-1])
-
-processC :: Signal Int -> Signal Int
-processC s_in = actor11SDF 1 1 f s_in where
-    f [a] = [a+1]
-
-processD :: Signal Int -> Signal Int -> Signal Int
```
klarasm, line 29: Do we want to add support for having the types specified in the CoreIR to ForSyDeIR translation? In that case that might be a good issue to create.

examples/test/double.hcr
klarasm: I think this should be resolved with #178.

src/CoreToForSyDeIR.hs
```diff
@@ -229,54 +257,82 @@ genSignalName context =
...
+                    [i] -> i
+                    _ -> error ("getSourceRate: More than one output rate for actor " ++ pcId)
+            else aux pcTail
 
+-- | Creates actors and delays by pattern matching based on the number of
+-- inputs to the top function, represented by `Lam`, and inputs to the process
+-- constructor, represented by `App`. If it cannot match an actor or delay then
+-- it creates a function.
 translateCoreExpr :: TranslationContext -> CoreBndr -> CoreExpr -> TranslationContext
 translateCoreExpr context binder expr = case expr of
```
klarasm, line 300: Eventually (probably not now) it might be nice to automatically construct the pattern matching from just the number of inputs and outputs, so we could provide a list, like [..., ("actor21SDF", 2, 1), ...] and it would match all of them for us.

klarasm: Related to this, it might be nice to have a comment on why we have the number of Lams and Apps. It would be nice to have a way of adding new ones without having to decipher the Core each time.

src/ForSyDeIR.hs
```diff
@@ -86,17 +86,23 @@ prettyIRConstructor (IRActor actorId actorType functionId (inputs, outputs)) =
...
+    functionId
+    (maybe "" (prettyFunction dflags) function)
+
+prettyFunction :: DynFlags -> CoreExpr -> String
+prettyFunction dflags function = printf "\n%s" (indent 2 (prettyCoreExpr dflags function))
 
 prettyIRSystem :: DynFlags -> IRSystem -> String
 prettyIRSystem dflags (IRSystem (inputs, outputs) constructors signals functions) =
   printf
     "IRSystem(\n  {%s}, {%s},\n  {\n%s  },\n  {\n%s  },\n  {\n%s  }\n)"
```
klarasm, line 100: This is where the final newline is missing, causing the mismatch between the test files.

klarasm (commented):
I think you did well with the comments now in general. It might be nice to have some more in `translateToCoreExpr`, at least the reasoning for the pattern matching being what they are (I think it would make it easier to reason about adding new cases).

SamuelMiksits (commented):
I think it is good enough now. The only issue is it might cause conflicts with #178 ?

klarasm:
> I think it is good enough now. The only issue is it might cause conflicts with #178 ?

I can fix that afterwards, feel free to merge on my part.

SamuelMiksits:
> > I think it is good enough now. The only issue is it might cause conflicts with #178 ?
> 
> I can fix that afterwards, feel free to merge on my part.

Okay just to clarify, do you want me to merge?

klarasm:
> > > I think it is good enough now. The only issue is it might cause conflicts with #178 ?
> > 
> > 
> > I can fix that afterwards, feel free to merge on my part.
> 
> Okay just to clarify, do you want me to merge?

Yes, I don't have any issues with this really, and I'll fix it in the newline stuff in my PR.

klarasm:
Btw @sthaeron, did you intend to define a new HSpec test for this as you added the `double.hcr`?

PR #178, opened 2025-11-06
Use POSIX newline convention consistently

klarasm:
This should make VSCode comply with POSIX convention.


klarasm:
(Note that this does address the JSON output test, as the `encode` function does not create a newline by default. I'll take a look at that later)

SamuelMiksits (commented):
I haven't tried running the tests (EDIT: Now I have ran the tests and they pass), but looking at the manual output I think this is good.
The only thing is whether you should do in `Main.hs` at L46:

```diff
- putStrLn s
+ putStr s
```

since currently I get the following output:

```
bash-5.3$ cabal exec forsyde-devtools-exe -- ./examples/model/SDF_example_001.hs --stdout
...
  ))
)

bash-5.3$ 
```
(an extra newline between end of the file and bash prompt when using `--stdout`)

klarasm:
> I haven't tried running the tests, but looking at the manual output I think this is good. The only thing is whether you should do in `Main.hs` at L46:
> 
> ```diff
> - putStrLn s
> + putStr s
> ```
> 
> since currently I get the following output:
> 
> ```
> bash-5.3$ cabal exec forsyde-devtools-exe -- ./examples/model/SDF_example_001.hs --stdout
> ...
>   ))
> )
> 
> bash-5.3$ 
> ```
> 
> (an extra newline between end of the file and bash prompt when using `--stdout`)

Right, forgot that. I'll have to handle the JSON too for that to work, but there seems there is a package which can do it (and make it a little prettier too without needing to pass it to `json_pp`).

klarasm:
@sthaeron I changed what we talked about earlier.

SamuelMiksits (approved):
Okay now I could build, the tests passed, and this time I got: 

```
bash-5.3$ cabal exec forsyde-devtools-exe -- ./examples/model/SDF_example_001.hs --stdout
...
)
bash-5.3$ 
```
so I am happy. There is also nothing that looked particularly egregious in the code.

klarasm:
Oops, I don't think I removed the trailing whitespace from the generated core output.

sthaeron (approved):
Looks good to me, tested it myself and everything works as expected.

PR #185, opened 2025-11-07
add test for single core static scheduler

Himawari722:
I have moved the test from the main in SDF scheduler to a new file with spec, and changed the original printing functions from printing to IO to output a string. Also I want to change the name of SDF_scheduler.hs to SDFScheduler.hs and corresponding module name to align with other files, but I keep that for now since I think it might make the git file compare messy. I can change the name once the current modification is reviewed and approved.


sthaeron (changes requested):
I don't think the tests for the scheduler should be based on the string output. Since, in this case we don't want to be testing the pretty printer but the actual output of the scheduler that the compiler will be using. Could you change the tests to use your `computeScheduleAndBuffers` function and just compare them to a pre calculated `([String], [Int])` rather than a `schedule*.txt` file.

sthaeron:
We should also add a backlog issue relating to adding a compiler argument for outputing information from the scheduler to standard out. It would be a good use of `computeScheduleAndBuffersPrint` and since we have the information might as well allow the user to access it if they want.

sthaeron:
When you change the name of `SDF_schedule.hs` to `SDFScheduler.hs` don't forget to change the test spec from `SDFScheduleSpec.hs` to `SDFSchedulerSpec.hs`.

examples/test/Scheduler/schedule6.txt
```diff
@@ -0,0 +1,44 @@
+Topology Matrix (Edges × Actors):
```
klarasm: I agree with @sthaeron that we should compare the expected data structure output rather than this text output. In the data structure we want to pass along much of the information here can probably dropped.

Perhapse something like this:
```haskell
data Schedule = SDFSchedule [String] [(String, Int)]
```
where the first parameter is the actor firing order and the second is the signal identifier and the buffer size of each signal. It would also be nice to have the number of tokens we need for the global input per run (maybe also global output) in the buffer sizes if that's possible.

examples/test/Scheduler/schedule6.txt
```diff
@@ -0,0 +1,44 @@
...
+Verification of null space vector:
+Topology Matrix × Repetition Vector ≈ Zero Vector? True
+
+Normalized repetition vector for ACTORS (integers):a=8
+b=2
+c=4
+d=1
+
+Generated schedule (actor firing order):
+d, c, a, a, c, a, a, b, c, a, a, c, a, a, b
```
klarasm: This should be part of the data structure which is passed along. Additionally, it would also be nice to have the number of tokens we need to process on the global inputs due to how we're doing that in the C implementation (where we get all input before we execute the schedule).

examples/test/Scheduler/schedule6.txt
```diff
@@ -0,0 +1,44 @@
...
+ c → a: 0
+ b → d: 2
+
+Verification of schedule with computed initial tokens: OK
+
+Simulated buffer sizes (maximum tokens observed per edge):
+  a → b: 4
+  d → c: 4
+  c → a: 4
+  b → d: 2
```
klarasm: This should be part of the data structure we pass along, ideally with the same signal name as specified in the ForSyDe IR.

Himawari722:
> I don't think the tests for the scheduler should be based on the string output. Since, in this case we don't want to be testing the pretty printer but the actual output of the scheduler that the compiler will be using. Could you change the tests to use your `computeScheduleAndBuffers` function and just compare them to a pre calculated `([String], [Int])` rather than a `schedule*.txt` file.

That makes sense — I initially used the string output since it already included the computed schedule and buffer sizes, so it indirectly tested the core functionality as well.
But I agree that comparing the raw result of computeScheduleAndBuffers to pre-calculated values would make the test more focused and less dependent on the pretty printer.
I’ll update the tests accordingly.

test/SDFScheduleSpec.hs
```diff
@@ -0,0 +1,157 @@
+module SDFScheduleSpec (spec) where
+
+import Data.Char (isSpace)
+import ForSyDeIR
+import SDF_schedule
+import Test.Hspec
+
+----------------------------------------------------------
+-- Example Systems to test the algorithm
+----------------------------------------------------------
```
klarasm, line 9: I think it's good that we have these hand-rolled examples here, but it might also be nice to have a few end-to-end examples which start from the ForSyDe haskell model. I think the Core to ForSyDe transformation should be able to handle all examples in the repo except for `SDF_exampe_001.hs`.

src/SDF_schedule.hs
klarasm: Just checking to see if I'm understanding correctly here, is the first element in the tuple the order the actors should fire (the actor names), and the second one the buffer sizes of the signals?

In that case it should be easy to reorganize into a data structure the one above. We just need the signal names.

klarasm (changes requested):
I would also like to have the tests compare the data structure instead of the string.

examples/test/Scheduler/schedule6.txt
Himawari722: I was actually wondering about whether we really need to compute the global input size. Although the current C implementation fetches all inputs before executing the schedule, it can lead to unnecessarily large buffers when an input actor fires multiple times. Fetching input only when the actor actually needs it might be more efficient (kind of like Haskell’s laziness, ha). But I can adjust it either way depending on what we decide.

test/SDFScheduleSpec.hs
Himawari722: Yes I can also add these to the Haskell examples. Also since I start working on ForSyDeIR(system part) to Procedural IR and the whole pipeline can work, maybe we can modify the current individual tests for each module to test the whole pipeline from the beginning

Himawari722:
I have updated the scheduler and corresponding test with the following changes:

1. **Enhanced Edge Data Structure**: Added `edgeName` field to preserve original signal names from IRSystem.

2. **Improved Buffer Size Reporting**: Modified scheduler to return buffer sizes as `[(String, Int)]` with corresponding signal names.

3. **Added I/O Buffer Calculation**: Implemented `computeIOBufferSizes` function to calculate buffer sizes for global input/output edges by calculating rate × repetition count.

4. **Fixed All Compiler Warnings**: Resolved naming conflicts, unused variables, and incomplete pattern matches.

**Example Output for exampleSystem4:**
**Schedule**: `["actor_a", "actor_a", "actor_b", "actor_c", "actor_d"]`

**Buffer Sizes:**

- I/O Buffers:
  - `s_ina: 4` (input rate 1 × actor_a repetition 2 × consumption rate 2)
  - `s_inb: 1` (input rate 1 × actor_b repetition 1 × consumption rate 1)
  - `s_out: 2` (actor_d repetition 1 × production rate 2)

Internal Buffers:

`s_1`: 2, `s_2`: 2, `s_3`: 1, `s_4_delay_s_4`: 1

Note that the named of delay edges are combined (`s_4` and `delay_s_4`) to represent the combined delay edges.

If the current version is nice enough, I can update the document and file name then we can merge it.

klarasm (approved):
I think all my comments have been addressed, sorry for taking a while to look at this again.

Himawari722:
@sthaeron Is it ok for me to merge this now?

klarasm:
I think `CoreToForSyDeIr.hs` was renamed and split into `CoreIRToForSyDeIR.hs` and `Utilities.hs`.

Himawari722:
> I think `CoreToForSyDeIr.hs` was renamed and split into `CoreIRToForSyDeIR.hs` and `Utilities.hs`.

Thank you so much for this!

sthaeron (approved):
Everything seemed to work when I tested it.

PR #187, opened 2025-11-07
examples: Add examples with more actor types such as actor14SDF

SamuelMiksits:
Since it is not possible to test more complicated actors I added 2 more:

- `model/SDF_example_006.hs` - Has no delays, but contains actors `actor14SDF`, `actor21SDF`, `actor12SDF`, `actor31SDF`. Functions contains only addition.
- `model/SDF_example_007.hs` - Has no delays, but contains actors `actor12SDF`, `actor13SDF`, `actor32SDF`, `actor41SDF`. Functions contains only addition.

In general I think we should start thinking about developing "coverage models" in the form of a table over the example models we have that displays which:

 - actor types
 - if they have delays
 - what functions they have (addition, subtraction, multiplication, division) 

so that it becomes clear what tests we are performing and which models need to be added for it to be possible to work on a specific feature (because obviously it is completely impossible to implement a feature if there is no example model for that in the repo! /s)

### DIFF
--- a/examples/model/SDF_example_006.hs
+++ b/examples/model/SDF_example_006.hs
@@ -1,0 +1,45 @@
+module SDF_example_006 where
+
+import ForSyDe.Shallow
+
+-- Netlist
+system :: Signal Int -> Signal Int
+system s_in = s_out
+  where
+    (s_1, s_2, s_3, s_4) = a_a s_in
+    s_5 = a_b s_1 s_2
+    (s_6, s_7) = a_c s_3
+    s_8 = a_d s_5 s_6
+    s_out = a_e s_8 s_7 s_4
+
+-- Process specifications
+a_a :: Signal Int -> (Signal Int, Signal Int, Signal Int, Signal Int)
+a_a s_1 = actor14SDF 2 (1, 3, 2, 4) f_1 s_1
+
+a_b :: Signal Int -> Signal Int -> Signal Int
+a_b s_1 s_2 = actor21SDF (1, 3) 2 f_2 s_1 s_2
+
+a_c :: Signal Int -> (Signal Int, Signal Int)
+a_c s_1 = actor12SDF 2 (2, 1) f_3 s_1
+
+a_d :: Signal Int -> Signal Int -> Signal Int
+a_d s_1 s_2 = actor21SDF (2, 2) 2 f_4 s_1 s_2
+
+a_e :: Signal Int -> Signal Int -> Signal Int -> Signal Int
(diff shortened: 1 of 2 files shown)


SamuelMiksits:
Hmmm, changed in `SDF_example_007.hs` from

```hs
f_2 [x] = ([x, x + 1], [1], [x + 2, x + 1])
```

to

```hs
f_2 [x] = ([x, x + 1], [x], [x + 2, x + 1])
```

Interestingly enough, the first case caused actor `a_b` to have:

```
NonRec(a_b =
  Lam(d ->
  Lam(c ->
  Lam($dNum ->
  Lam($dNum ->
  Lam(s_1 ->
  App(App(App(App(App(App(App(App(Var(actor13SDF) *
```
and the second one got:

```
NonRec(a_b =
  Lam(d ->
  Lam($dNum ->
  Lam(s_1 ->
  App(App(App(App(App(App(App(App(Var(actor13SDF) *
```

Is this a problem? I don't think it makes sense to use a function that is hardwired to a constant but it might be interesting to be aware/explore this topic.

sthaeron (changes requested):
Great initiative, having more complex net list examples will help facilitate scaling the Core IR to ForSyDe IR translation. I have updated the translation solution in the pr #194 to require type declarations for the net list, process constructors, and functions. Could you update your new examples to include type declarations?

SamuelMiksits:
> Great initiative, having more complex net list examples will help facilitate scaling the Core IR to ForSyDe IR translation. I have updated the translation solution in the pr #194 to require type declarations for the net list, process constructors, and functions. Could you update your new examples to include type declarations?

Sure! I was taking a look at the PR. I had actually implemented every remaining actor locally but I guess all I have to do is just remove 2 `Lam(` from every line and the closing parentheses. I ran into a brick wall with trying to debug the code for my new models since I am so used to using GDB or adding print statements, neither of which is really an option in Haskell but I was getting something ridiculous  where a actor14SDF with `s_in` would get `s_1` instead as input and have 6 outputs, and none of the other actors had any signals

sthaeron:
> I was getting something ridiculous where a actor14SDF with `s_in` would get `s_1` instead as input and have 6 outputs, and none of the other actors had any signals

I reimplemented the solution for identifying how signals relate to the inputs and outputs of process constructors in the new core transformation. Hopefully, it'll resolve your issues. Otherwise it might be that how we pattern match on creating actors is not robust enough. We should probably be pattern matching on the `Var` type which has the corresponding actor string rather than the  actor string itself.

SamuelMiksits:
Okay, rebased on current dev and added types

sthaeron (approved):
Looks good to me!

PR #194, opened 2025-11-09
Core output now has no inlining!

sthaeron:
- Made GHC typechecked module to not inline ids within the "system" function. Solution is based on discussions from: https://gitlab.haskell.org/ghc/ghc/-/issues/24386.
- Updated translation from Core IR to ForSyDe IR to take advantage of the new no inlined Core IR system net list. I think the new translation solution is much simpler to scale for more complex net lists and easier to understand.
- Added a much more robust solution for updating the inputs and outputs of constructors, no longer using a hacky solution.
- Updated comments for the new translation solution.
- Updated SDF examples to have types, it made the Core pattern matching much simpler.

Note currently SDF examples 3 is supported, as the goal was to update our solution be based on no inlined Core

### DIFF
--- a/examples/model/SDF_example_002.hs
+++ b/examples/model/SDF_example_002.hs
@@ -3,6 +3,7 @@ module SDF_example_002 where
 import ForSyDe.Shallow
 
 -- Netlist
+system :: Signal Int -> Signal Int -> Signal Int
 system s_in_1 s_in_2 = s_out
   where
     s_1 = a_a s_in_1
@@ -12,21 +13,30 @@ system s_in_1 s_in_2 = s_out
     s_4_delayed = d_1 s_4
 
 -- Process specifications
+a_a :: Signal Int -> Signal Int
 a_a s_1 = actor11SDF 2 1 f_1 s_1
 
+a_b :: Signal Int -> Signal Int
 a_b s_1 = actor11SDF 1 2 f_2 s_1
 
+a_c :: Signal Int -> Signal Int -> Signal Int
 a_c s_1 s_2 = actor21SDF (2, 1) 1 f_3 s_1 s_2
 
+a_d :: Signal Int -> Signal Int -> (Signal Int, Signal Int)
 a_d s_1 s_2 = actor22SDF (2, 1) (2, 1) f_4 s_1 s_2
 
+d_1 :: Signal Int -> Signal Int
 d_1 s = delaySDF [0] s
 
 -- Function definitions
+f_1 :: [Int] -> [Int]
 f_1 [x, y] = [x + y]
 
+f_2 :: [Int] -> [Int]
 f_2 [x] = [x, x + 1]
 
+f_3 :: [Int] -> [Int] -> [Int]
 f_3 [x, y] [z] = [x + y + z]
 
+f_4 :: [Int] -> [Int] -> ([Int], [Int])
 f_4 [x, y] [z] = ([x + y + z], [x + y, x + y + z])
--- a/examples/model/SDF_example_002.hs
+++ b/examples/model/SDF_example_002.hs
@@ -9,7 +9,7 @@ system s_in_1 s_in_2 = s_out
     s_1 = a_a s_in_1
     s_2 = a_b s_in_2
     s_3 = a_c s_1 s_4_delayed
-    (s_out, s_4) = a_d s_2 s_3
+    (s_4, s_out) = a_d s_2 s_3
     s_4_delayed = d_1 s_4
 
 -- Process specifications
@@ -23,7 +23,7 @@ a_c :: Signal Int -> Signal Int -> Signal Int
 a_c s_1 s_2 = actor21SDF (2, 1) 1 f_3 s_1 s_2
 
 a_d :: Signal Int -> Signal Int -> (Signal Int, Signal Int)
-a_d s_1 s_2 = actor22SDF (2, 1) (2, 1) f_4 s_1 s_2
+a_d s_1 s_2 = actor22SDF (2, 1) (1, 2) f_4 s_1 s_2
 
 d_1 :: Signal Int -> Signal Int
 d_1 s = delaySDF [0] s
--- a/forsyde-devtools.cabal
+++ b/forsyde-devtools.cabal
@@ -55,7 +55,7 @@ executable forsyde-devtools-exe
         CoreIR
         ForSyDeIR
         ProceduralIR
-        CoreToForSyDeIR
+        CoreIRToForSyDeIR
         ProceduralIRToC
         Utilities
 
--- a/forsyde-devtools.cabal
+++ b/forsyde-devtools.cabal
@@ -28,7 +28,7 @@ common common-options
         bytestring >=0.11 && <0.13,
         process == 1.6.25.0,
         hmatrix == 0.20.2,
-        syb == 0.7.3,
+        syb >= 0.7.2.4 && <= 0.7.3,
         lsp == 2.7.0.1,
         network == 3.2.7.0,
         co-log-core == 0.3.2.5,
@@ -66,8 +66,11 @@ executable forsyde-lsp-exe
     ghc-options:      -rtsopts -with-rtsopts=-N
     build-depends:    forsyde-devtools
     other-modules:
-        SKGraphSchema
         ArgumentsLSP
+        ForSyDeIR
+        CoreIRToForSyDeIR
+        SKGraphSchema
+        Utilities
 
 test-suite forsyde-devtools-test
     import:           common-options
--- a/forsyde-devtools.cabal
+++ b/forsyde-devtools.cabal
@@ -67,10 +67,11 @@ executable forsyde-lsp-exe
     build-depends:    forsyde-devtools
     other-modules:
         ArgumentsLSP
+        CoreIR
         ForSyDeIR
         CoreIRToForSyDeIR
-        SKGraphSchema
         Utilities
+        SKGraphSchema
 
 test-suite forsyde-devtools-test
     import:           common-options
--- a/forsyde-devtools.cabal
+++ b/forsyde-devtools.cabal
@@ -28,6 +28,7 @@ common common-options
         bytestring == 0.12.2.0,
         process == 1.6.25.0,
         hmatrix == 0.20.2,
+        syb == 0.7.3,
         lsp == 2.7.0.1,
         network == 3.2.7.0,
         co-log-core == 0.3.2.5,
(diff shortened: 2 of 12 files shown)


sthaeron:
Updated new core translation solution to support SDF examples 1 to 5. They all seem to be working, I have only hand checked SDF example 3. For the rest I've only checked that the process constructors and their local inputs and outputs are correct.

klarasm:
I just merged the prototype LSP so there are some conflicts now, sorry about that. Hopefully they're easy to resolve.

I'll run this for a test.

forsyde-devtools.cabal
```diff
@@ -27,6 +27,8 @@ common common-options
         containers == 0.7,
         bytestring == 0.12.2.0,
         hmatrix == 0.20.2,
+        syb == 0.7.3,
```
klarasm: As the CI run tells you, this dependency does not work out. If I change it to 0.7.2.4 it does.

examples/model/SDF_example_002.hs
```diff
@@ -3,30 +3,40 @@ module SDF_example_002 where
...
 
 -- Netlist
+system :: Signal Int -> Signal Int -> Signal Int
 system s_in_1 s_in_2 = s_out
   where
     s_1 = a_a s_in_1
     s_2 = a_b s_in_2
     s_3 = a_c s_1 s_4_delayed
-    (s_out, s_4) = a_d s_2 s_3
+    (s_4, s_out) = a_d s_2 s_3
```
klarasm, line 12: Was this change intentional?

forsyde-devtools.cabal
```diff
@@ -45,7 +47,8 @@ executable forsyde-devtools-exe
         Arguments
         CoreIR
         ForSyDeIR
-        CoreToForSyDeIR
+        CoreIRToForSyDeIR
+        Utilities
```
klarasm, line 60: I needed these changes for the LSP to work (I probably should have added the other-modules initially though):

```diff
diff --git a/forsyde-devtools.cabal b/forsyde-devtools.cabal
index 0a6039c09841..f7ba23462f56 100644
--- a/forsyde-devtools.cabal
+++ b/forsyde-devtools.cabal
@@ -64,6 +64,9 @@ executable forsyde-lsp-exe
     other-modules:
         SKGraphSchema
         ArgumentsLSP
+        ForSyDeIR
+        CoreIRToForSyDeIR
+        Utilities
 
 test-suite forsyde-devtools-test
     import:           common-options
diff --git a/src/LSP.hs b/src/LSP.hs
index 3816f7090cf2..eb0292ea998f 100644
--- a/src/LSP.hs
+++ b/src/LSP.hs
@@ -14,8 +14,7 @@ import qualified Control.Exception as E
 import Control.Monad (forever, void)
 import Control.Monad.IO.Class
 import Control.Monad.IO.Unlift
-import CoreIR
-import CoreToForSyDeIR
+import CoreIRToForSyDeIR
 import Data.Aeson ((.=))
 import qualified Data.Aeson as A
 import Data.Aeson.KeyMap ((!?))
@@ -31,6 +30,7 @@ import Options.Applicative
 import Prettyprinter
 import SKGraphSchema
 import System.IO
+import Utilities
 
 diagramAcceptMethod :: SMethod (Method_CustomMethod "diagram/accept")
 diagramAcceptMethod = (SMethod_CustomMethod (Proxy @"diagram/accept"))
```

sthaeron:
> merged the prototype LSP

No worries, the conflicts were easy to resolve. You should now be able to test the LSP prototype from haskell code to diagram on this branch for all five SDF examples.

src/Utilities.hs
```diff
@@ -0,0 +1,85 @@
...
+          }
+  _ <- setSessionDynFlags newDflags
+  target <- guessTarget filePath Nothing Nothing
+  setTargets [target]
+  _ <- load LoadAllTargets
+  modSummary <- getModSummary $ mkModuleName (takeBaseName filePath)
+  env <- getSession
+  parsedModule <- liftIO $ hscParse env modSummary
+  (tcg, _) <- liftIO $ hscTypecheckRename env modSummary parsedModule
+  let noInlineTcg = noInlineTypecheck tcg
```
klarasm, line 34: Oh! So this just applies as an intermediate step in the compilation process, no need for template haskell. Nice!

src/Utilities.hs
```diff
@@ -0,0 +1,85 @@
...
+  guts <- liftIO $ hscDesugar env modSummary noInlineTcg
+  return $ (mg_binds guts, newDflags)
+
+-- | Updates all bindings within the function called `system` and adds NOINLINE
+-- pragmas. Prevents the pre optimisier run during desurgaring from inlining
```
klarasm: "desurgaring" -> desugaring.

klarasm (changes requested):
Nice! Tried this on all the examples for the LSP and it seems reasonable. However, now we should probably suppress the edge label from/to global inputs/outputs as they are the same as the node name.

Some nitpicks and comments below.

examples/model/SDF_example_002.hs
sthaeron: Yes, both the output rates for actor `a_d` and the outputs was changed in an earlier pr and now I'm changing it back. Had changed it previously since the old solution for associating signals to the outputs of process constructors was a bit hacky and required system outputs to always be listed first. However, the new solution is now able to correctly identify the order the outputs of a process constructor are in.

klarasm: This would be nice to have in a separate commit with an explanation if it's not too much effort, then.

sthaeron:
> suppress the edge label from/to global inputs/outputs as they are the same as the node name.

Yes, I think this is a better solution than potentially making new names for either the signals or global system inputs/outputs.

src/Utilities.hs
sthaeron: Yes and it only no inlines the ids of nodes within the net list's function called "system".

forsyde-devtools.cabal
sthaeron: I initially used 0.7.3 for `syb` as it was the only listed version that had been tested with ghc version `9.10`. Pushed an update to include 0.7.2.4, hopefully it works.

forsyde-devtools.cabal
sthaeron: Just added your changes in a commit.

forsyde-devtools.cabal
```diff
@@ -59,8 +61,11 @@ executable forsyde-lsp-exe
     ghc-options:      -rtsopts -with-rtsopts=-N
     build-depends:    forsyde-devtools
     other-modules:
-        SKGraphSchema
         ArgumentsLSP
+        ForSyDeIR
```
SamuelMiksits, line 71: I get:

```
<no location info>: warning: [GHC-32850] [-Wmissing-home-modules]
    These modules are needed for compilation but not listed in your .cabal file's other-modules for ‘forsyde-devtools-0.0.0.1-inplace-forsyde-lsp-exe’ :
        CoreIR
```

and this goes away if we add CoreIR as dependency here. Should we do this? I haven't tested if it works or not

src/CoreIRToForSyDeIR.hs
```diff
@@ -0,0 +1,404 @@
...
+     in translateSystemBinds context2 bindTail
+
+-- | Translates system `CoreExpr`. Identifies if the expression represents an
+-- application of a process constructor or connection to a specific output of a
+-- process constructor. In either case returns a `Binder` along side a
+-- potentially updated `TranslationContext`.
+--
+-- NOTE: This function needs to be updated with new pattern matches for
+-- translation to support additional process constructors.
+translateSystemExpr :: TranslationContext -> CoreExpr -> (Binder, TranslationContext)
```
SamuelMiksits, line 210: So I ran the LSP on `SDF_example_002.hs` and it works, but this function only pattern matches on `delaySDF` and `actor22SDF` (and binding). That model obviously has other as well (`actor11SDF`, `actor21SDF`, noteworthy with only one output), so I don't quite understand when this is called (possibly only with multiple outputs, but why then is delay here), and I am running into issues with my other actors but those actors still work. Do we need to add `actor11SDF`/`actor21SDF` later?

SamuelMiksits (approved):
I think overall it is fine (ready to merge), but I am wondering a bit about the `CoreIR` dependency compiler warning and process constructors in `translateSystemExpr`.

SamuelMiksits:
By the way the title "Core output is now has no inlining!" makes no sense. Is it supposed to say "Core output now has no inlining!"?

src/CoreIRToForSyDeIR.hs
sthaeron: I completely forgot about the applications of `actor11SDF`, `actor12SDF`, and `actor21SDF`. The reason that it does work turns out to be because the pattern matching for the application of process constructors is the same if they have the same number of inputs. Meaning the current `translateSystemExpr` should work for process constroctors with 1 inputs such as `delaySDF` and `actor1*SDF` or with 2 inputs `actor2*SDF`.

forsyde-devtools.cabal
sthaeron: If the compiler complains we probably should add it. I added the other modules based on Klara's suggestion, I haven't tested the LSP exe myself.

examples/model/SDF_example_002.hs
sthaeron: Not sure how to accomplish this. I sneakily changed it in a previous pr and @SamuelMiksits was confused by it. My intention was to sneakily change it back to how it was before, as the initial change was just a crutch for the old hacky solution to work.

klarasm: I usually do something along the lines of:
Start an interactive rebase and select edit on the offending commit. Then do `git revert -n HEAD`. This stages a revert but does not commit it. Now, `git stash -p` the changes you want to keep in the commit and then `git add -u`. Commit the changes and now do `git revert HEAD`. Now, you can either complete the rebase and start another or do `git rebase --edit-todo`. Squash the first commit into the previous one and reword the revert of it to something proper.

klarasm: The resulting commit is empty it seems and the changes are in the previous one.

sthaeron: I think I squashed the wrong commits together. I had done the following:
```
pick * #  feat: translate core accepts SDF examples 1 to 5
pick * # reverted change
squash * # Revert "reverted change"
```

klarasm: Ah, yes. It's the second that should be squashed with the first.

klarasm (approved):
Looks good now!

PR #196, opened 2025-11-12
Add KArc and KSpline

SamuelMiksits:
Added `KArc` and `KSpline` I just tested that they worked by replacing 
```hs
        edge =
          KEdge
            { gid = name,
              children = [KLabel {gid = sigid, label = T.pack n}],
              renderings = [KRoundedBendsPolyline [] 4],
              properties = [],
              source = sn,
              target = tn
            }
```
in LSP with the different `renderings = ` for the other types of lines. for `example/models/SDF_example_003.hs`. `KArc` did not really work that well since hardcoding all edges to have the same start angle and arc angle doesn't really work.

### DIFF
--- a/src/SKGraphSchema.hs
+++ b/src/SKGraphSchema.hs
@@ -89,7 +89,9 @@ instance A.ToJSON KStyle where
 data KRendering
   = KEllipse [KStyle]
   | KPolyline [KStyle]
-  | KRoundedBendsPolyline [KStyle] Int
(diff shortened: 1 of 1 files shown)


klarasm:
Looks good!
The arc is not really suitable for edges, it's not an arc in the edge sense, rather it's a Circular Arc.

klarasm (approved):

Issue #197, opened 2025-11-12
Input/Output signal edges should have no label


klarasm:
Now that we have no inlining for the net list, the label of the input/output nodes is the same as the signal.

PR #241, opened 2025-11-12
print edge labels only when name is the different from source/target

klarasm:


SamuelMiksits (commented):
I think the code looks fine. But I don't really understand what the purpose is of this patch. What scenario produces a graph that is wrong, what does it look like and why is it wrong?

(EDIT: Ah is this related to #197 ?)

(EDIT2: Yes it looks like it, now I don't see the s_in,s_out twice so it is good)

SamuelMiksits:
Okay I will just merge it now

PR #254, opened 2025-11-15
feat: added more globals to Procedural IR

sthaeron:
Add the rest of the globals, added type qualifiers, and storage classes

### DIFF
--- a/src/ProceduralIR.hs
+++ b/src/ProceduralIR.hs
@@ -60,6 +60,7 @@ data BinaryOperator
   | LessEqual -- lhs <= rhs
   | Greater -- lhs > rhs
   | GreaterEqual -- lhs >= rhs
+  deriving (Show)
 
 -- Types
 data Type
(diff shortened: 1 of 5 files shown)


src/ProceduralIR.hs
```diff
@@ -91,7 +109,11 @@ data Statement
 
 -- Globals
 data Global
-  = GFuncDef [String] Type String [(Type, String)] Statement
+  = GFuncDeclare (Maybe StorageClass) Type String [(Type, String)]
+  | GFuncDef (Maybe StorageClass) Type String [(Type, String)] Statement
```
Himawari722, line 115: I think function could have multiple qualifiers (`static const`) instead of just one storage class here?

src/ProceduralIR.hs
```diff
@@ -91,7 +109,11 @@ data Statement
...
+  | GVarDeclare Type String
+  | GVarDef Type String Expression
```
Himawari722, line 117: I think a list of `TypeQualifier` could also be added here as well since you already introduced and defined it above

Himawari722 (changes requested):
The new ProceduralIR is very nice! I only have a design concern before merging. 

For the GFuncDeclare and GFuncDefs, it currently use `Maybe StorageClass` which means a function can have at most one storage-class specifier, but I think there can be multiple specifiers from both storage class and type qualifiers such as `static const` or `static inline`. Also for global variables, a list of qualifiers could be added.

MichelD-Abo (approved):
Looks Great!

PR #257, opened 2025-11-18
Add KPolygon and other required types

SamuelMiksits:
Adds KPolygon and some types that it requires. This also implements arrows.
To test, go to `src/LSP.hs` L106 and:

```diff
- outports = map (createPort nid (maybe [] (\_l -> [KText "◆" []]) l)) outsignals
+ outports = map (createPort nid (maybe [] (\_l -> [kArrow]) l)) outsignals
```

The implementation has the issues of arrows getting resized, and the arrows do not automatically rotate to face the correct direction, (for instance `/examples/models/SDF_example_002.hs` shows this).

### DIFF
--- a/src/SKGraphSchema.hs
+++ b/src/SKGraphSchema.hs
@@ -50,6 +50,24 @@ data KProperty
   | EdgeType
   | JunctionPoints
 
+data KPlacementData
+  = KTopPosition Float Float -- absolute (Float), relative (Float)
+  | KLeftPosition Float Float -- absolute (Float), relative (Float)
+  | KRightPosition Float Float -- absolute (Float), relative (Float)
+  | KBottomPosition Float Float -- absolute (Float), relative (Float)
+
+-- "Fake" KRendering called "KArrow". A function that returns a KPolygon that
+-- creates an arrow. KLineJoin rounds the edges on the arrow
+kArrow :: KRendering
+kArrow =
+  KPolygon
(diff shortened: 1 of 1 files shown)


src/SKGraphSchema.hs
```diff
@@ -95,6 +130,30 @@ data KRendering
   | KRectangle [KStyle]
   | KRoundedRectangle [KStyle] Float Float -- cornerWidth (Float), cornerHeight (Float)
   | KText T.Text [KStyle]
+  | KPolygon [KStyle] [(KPlacementData, KPlacementData)] -- points, list (x,y) [(KPlacementData, KPlacementData)]
+
+buildCoords placements = case placements of
```
klarasm, line 136: No type signature for this one.

klarasm (commented):
I think it looks good except the missing type signature.

src/SKGraphSchema.hs
SamuelMiksits: Okay I added it now. The compiler suggested
```Haskell
buildCoords :: (A.ToJSON v1, A.ToJSON v2) => [(v1, v2)] -> [A.Value]
```
but this doesn't really make sense since I only care about what happens when I am using `KPlacementData` and not any arbitrary type.

(EDIT: Weird, on the "Conversation" page it shows the outdated version and doesn't say "Outdated", but if I go to "Files changed" I see the new force pushed version)

klarasm (approved):

PR #258, opened 2025-11-18
Stdout for LSP argument parsing

### LINKED ISSUE #243: Argument parsing option for Stdin/Stdout in LSP
The VSCode extension might not use the same approach as the KLighD CLI. It would be good to support both TCP and Stdin/Stdout.

SamuelMiksits:
Created argument parsing for `stdin`/`stdout` for the LSP. Marked as draft/WIP since I am currently not 100% sure on how it is supposed to be integrated with the rest of the program. 

- If I have `--stdin` where do I need to `read`, since the function itself makes the function call return `-> IO ()` or `-> IO [Char]` or something so it needs to be wrapped at a point like `main`/`run` where IO is already performed. 

- Then if `--stdout` is passed it needs to be implemented such that the TCP server is not ran at all or something.

### DIFF
--- a/src/LSP.hs
+++ b/src/LSP.hs
@@ -17,7 +17,10 @@ import Control.Monad.IO.Unlift
 import CoreIRToForSyDeIR
 import Data.Aeson ((.=))
 import qualified Data.Aeson as A
+import Data.Aeson.Encode.Pretty as AP
 import Data.Aeson.KeyMap ((!?))
+import qualified Data.ByteString as BS
+import qualified Data.ByteString.Lazy.Char8 as BSL8
 import qualified Data.List.NonEmpty as NE
 import Data.Proxy
 import qualified Data.Sequence as Seq
@@ -244,7 +247,7 @@ runServerC =
     (L.cmap (fmap $ T.pack . show . pretty) (L.cmap show L.logStringStderr))
 
 -- | Process arguments for the LSP and run it
-main :: IO Int
+main :: IO ()
 main = run =<< execParser opts
   where
     opts =
@@ -256,23 +259,34 @@ main = run =<< execParser opts
         )
 
 -- | Start the LSP
-run :: Arguments -> IO Int
-run (Arguments (Host ip) (TCP p) f) =
-  runTCPServer (Just ip) p lsp
-  where
-    lsp s = do
-      handle <- socketToHandle s ReadWriteMode
-      runServerC handle handle $
-        ServerDefinition
-          { parseConfig = const $ const $ Right Nothing,
-            onConfigChange = const $ pure (),
-            defaultConfig = Nothing,
-            configSection = "demo",
-            doInitialize = \env _req -> pure $ Right env,
-            staticHandlers = \_caps -> (handlers f),
-            interpretHandler = \env -> Iso (runLspT env) liftIO,
-            options = defaultOptions
-          }
+run :: Arguments -> IO ()
+run (Arguments (Host ip) (TCP p) i_f) =
+  case i_f of
+    FromClient ->
+      runTCPServer (Just ip) p lsp
+      where
+        lsp s = do
+          handle <- socketToHandle s ReadWriteMode
+          -- server returns IO Int, wrapper with "pure ()" so that expression
+          -- returns IO ()
+          _ <-
+            runServerC handle handle $
+              ServerDefinition
+                { parseConfig = const $ const $ Right Nothing,
+                  onConfigChange = const $ pure (),
+                  defaultConfig = Nothing,
+                  configSection = "demo",
+                  doInitialize = \env _req -> pure $ Right env,
+                  staticHandlers = \_caps -> (handlers FromClient),
+                  interpretHandler = \env -> Iso (runLspT env) liftIO,
+                  options = defaultOptions
+                }
+          pure ()
+    InputFile f -> do
+      (core, dflags) <- withRunInIO (\_u -> compileToCore f)
+      let ir = translateCoreProgram dflags core
+      let graphMessage = requestBounds f ir
+      BSL8.putStrLn $ AP.encodePretty graphMessage
 
 -- | Listen on host and port, as well as accept and fork off connections
 runTCPServer :: Maybe HostName -> ServiceName -> (Socket -> IO a1) -> IO a2
--- a/src/ArgumentsLSP.hs
+++ b/src/ArgumentsLSP.hs
@@ -7,6 +7,10 @@ data Input
   = InputFile FilePath
   | FromClient
 
+data Output
+  = ToClient
+  | StdOut
+
 -- Host IP(V4) address
 data IP
   = Host String
@@ -26,6 +30,11 @@ data Arguments = Arguments
     input :: Input
   }
 
+inputTop :: Parser Input
+inputTop =
+  inputFile
+    <|> inputFromClient
+
 -- Handle file input. Default is to get the file from your client, but
 -- allow user to pass a filename directly to server for debugging.
 inputFile :: Parser Input
@@ -35,6 +44,14 @@ inputFile =
     ( metavar "INPUT"
         <> value FromClient
         <> help "Input filename (debug option), if unset get file from client"
+    )
+
+inputFromClient :: Parser Input
+inputFromClient =
+  flag'
+    FromClient
+    ( long "input-client"
+        <> help "Input from client (default)"
     )
 
 -- IP - String "127.0.0.1" by default, otherwise provide -a or --address and
@@ -70,4 +87,4 @@ arguments =
   Arguments
     <$> ipAddress
     <*> tcpPort
-    <*> inputFile
+    <*> inputTop


klarasm:
Ah sorry, I was probably unclear in what I wanted. Regardless of if this is the klighd-vscode or klighd-cli we will always (to my knowledge) get the sourceUri from the client. What I was thinking of when writing the issue was that we might do the LSP client/server communication through stdin/stdout (this to me seems to be the way LSP usually works, though @ramadhanafif probably knows more about that, please correct me if I'm wrong).

ramadhanafif:
So LSP can operate both in socket or stdin/stdout. We just need to agree on which one we are using, and ignore the rest. In the LSP Client, this will be also a trivial thing to change (between socket OR stdin/out).

On that regard, the stdin/out programme is a long-running ones, so not run-and-done kind of thing.

SamuelMiksits:
Okay so do we actually need this now? I can see a use case for `--stdout` regardless since ForSyDe-IR-JSON is different from the raw JSON output that we have here, so we could potentially have `--stdout` just be a "single run" that dumps the raw JSON. I haven't really figured out a good way to sniff the packet (I have tried tcpdump and listened on port 5007 and 7000 (which is what is displayed in the browser for the client) but it never shows any packets)

klarasm:
> Okay so do we actually need this now? I can see a use case for `--stdout` regardless since ForSyDe-IR-JSON is different from the raw JSON output that we have here, so we could potentially have `--stdout` just be a "single run" that dumps the raw JSON. I haven't really figured out a good way to sniff the packet (I have tried tcpdump and listened on port 5007 and 7000 (which is what is displayed in the browser for the client) but it never shows any packets)

I use `tcpdump -nvvAi lo port 5007`, but you can print out the message relatively easy with:
```diff
diff --git a/src/LSP.hs b/src/LSP.hs
index 6f0668e5732a..11719eaad5e4 100644
--- a/src/LSP.hs
+++ b/src/LSP.hs
@@ -18,6 +18,7 @@ import CoreIRToForSyDeIR
 import Data.Aeson ((.=))
 import qualified Data.Aeson as A
 import Data.Aeson.KeyMap ((!?))
+import qualified Data.ByteString.Lazy as BSL
 import qualified Data.List.NonEmpty as NE
 import Data.Proxy
 import qualified Data.Sequence as Seq
@@ -216,6 +217,7 @@ handlers f =
         (core, dflags) <- withRunInIO (\_u -> compileToCore file)
         let ir = translateCoreProgram dflags core
         let graphMessage = requestBounds file ir
+        _ <- withRunInIO (\_u -> BSL.putStr $ A.encode graphMessage)
         sendNotification diagramAcceptMethod graphMessage
         pure ()
     ]
```

EDIT: But the single run stdout would maybe be nice to have

SamuelMiksits:
> EDIT: But the single run stdout would maybe be nice to have

Okay I am working on adding single run to `stdout`. Can we make it so that sending an input file as argument automatically puts it to `stdout`? There is some really stupid behavior right now. If you run server with input file `forsyde-lsp-exe ./examples/model/SDF_example_002.hs` and then run the client with `./klighd-linux --ls_port 5007 ./examples/model/SDF_example_010.hs` then it will show the `SDF_example_002.hs` model but have `SDF_example_010.hs` in the URI. Then I can remove the output data type and only have input, but do we also want so you can use either `StdIn` or `InputFile` and in both cases it will output to `stdout`?

I am also a bit unhappy with how main is currently. Normally, you want to have `main :: IO()`, but since the `runServerC` is `-> IO Int` everything is `-> IO Int`. So if I try to print my message then that becomes `IO ()`. Right now I am trying to handle this by converting the server code from to

```hs
runServerC handle handle $
  ServerDefinition
    { parseConfig = const $ const $ Right Nothing,
      onConfigChange = const $ pure (),
      defaultConfig = Nothing,
      configSection = "demo",
      doInitialize = \env _req -> pure $ Right env,
      staticHandlers = \_caps -> (handlers i_f),
      interpretHandler = \env -> Iso (runLspT env) liftIO,
      options = defaultOptions
    }
```

to

```hs
x <- runServerC handle handle $
  ServerDefinition
    { parseConfig = const $ const $ Right Nothing,
      onConfigChange = const $ pure (),
      defaultConfig = Nothing,
      configSection = "demo",
      doInitialize = \env _req -> pure $ Right env,
      staticHandlers = \_caps -> (handlers i_f),
      interpretHandler = \env -> Iso (runLspT env) liftIO,
      options = defaultOptions
    }
case x of
  _ -> print ""
```
but this is pretty terrible. I need something to unwrap the `Int` part of `IO Int` to be able to do it

klarasm:
> > EDIT: But the single run stdout would maybe be nice to have
> 
> Okay I am working on adding single run to `stdout`. Can we make it so that sending an input file as argument automatically puts it to `stdout`? There is some really stupid behavior right now. If you run server with input file `forsyde-lsp-exe ./examples/model/SDF_example_002.hs` and then run the client with `./klighd-linux --ls_port 5007 ./examples/model/SDF_example_010.hs` then it will show the `SDF_example_002.hs` model but have `SDF_example_010.hs` in the URI. Then I can remove the output data type and only have input, but do we also want so you can use either `StdIn` or `InputFile` and in both cases it will output to `stdout`?
> 

That makes sense. We only added the specified file on the server for debugging anyway, so it would make sense for that to imply single shot stdout as well.

> I am also a bit unhappy with how main is currently. Normally, you want to have `main :: IO()`, but since the `runServerC` is `-> IO Int` everything is `-> IO Int`. So if I try to print my message then that becomes `IO ()`. Right now I am trying to handle this by converting the server code from to
> 
> ```haskell
> runServerC handle handle $
>   ServerDefinition
>     { parseConfig = const $ const $ Right Nothing,
>       onConfigChange = const $ pure (),
>       defaultConfig = Nothing,
>       configSection = "demo",
>       doInitialize = \env _req -> pure $ Right env,
>       staticHandlers = \_caps -> (handlers i_f),
>       interpretHandler = \env -> Iso (runLspT env) liftIO,
>       options = defaultOptions
>     }
> ```
> 
> to
> 
> ```haskell
> x <- runServerC handle handle $
>   ServerDefinition
>     { parseConfig = const $ const $ Right Nothing,
>       onConfigChange = const $ pure (),
>       defaultConfig = Nothing,
>       configSection = "demo",
>       doInitialize = \env _req -> pure $ Right env,
>       staticHandlers = \_caps -> (handlers i_f),
>       interpretHandler = \env -> Iso (runLspT env) liftIO,
>       options = defaultOptions
>     }
> case x of
>   _ -> print ""
> ```
> 
> but this is pretty terrible. I need something to unwrap the `Int` part of `IO Int` to be able to do it

maybe you could end it with `pure ()` instead of a print?

SamuelMiksits:
> maybe you could end it with `pure ()` instead of a print?

Ah thank you `pure ()` works

SamuelMiksits:
Okay I just added so that you can dump output to `stdout`. Usually advisable to pipe into a file or something since JSON output tends to be >1000 lines which would get lost in "terminal history". I didn't add `stdin` since it seemed to not really be possible to run GHC with stdin as input. I saw some conflicting information saying it was not possible with GHC (maybe with some plugin) but some other said it was possible specifically with `runghc`

src/LSP.hs
```diff
@@ -17,11 +17,15 @@ import Control.Monad.IO.Unlift
 import CoreIRToForSyDeIR
 import Data.Aeson ((.=))
 import qualified Data.Aeson as A
+import Data.Aeson.Encode.Pretty
```
klarasm: Might be nice to qualify this import too, you're only using a single function from it right?

src/LSP.hs
```diff
@@ -256,23 +260,34 @@ main = run =<< execParser opts
...
+                  staticHandlers = \_caps -> (handlers FromClient),
+                  interpretHandler = \env -> Iso (runLspT env) liftIO,
+                  options = defaultOptions
+                }
+          pure ()
+    InputFile f -> do
+      (core, dflags) <- withRunInIO (\_u -> compileToCore f)
+      let ir = translateCoreProgram dflags core
+      let graphMessage = requestBounds f ir
+      T.IO.putStrLn $ T.E.decodeUtf8 $ BS.toStrict $ encodePretty graphMessage
```
klarasm: Doesn't ByteString have a putStr too?

klarasm (approved):
Looks good! I have some small comments but don't really feel that strongly about them, so if you don't agree you can just merge as is.

src/LSP.hs
SamuelMiksits:
```
    • Couldn't match expected type ‘BS.ByteString’
                  with actual type ‘T.Text’
    • In the second argument of ‘($)’, namely
        ‘T.E.decodeUtf8 $ BS.toStrict $ encodePretty graphMessage’
      In a stmt of a 'do' block:
        BS.putStr
          $ T.E.decodeUtf8 $ BS.toStrict $ encodePretty graphMessage
      In the expression:
        do (core, dflags) <- withRunInIO (\ _u -> compileToCore f)
           let ir = translateCoreProgram dflags core
           let graphMessage = requestBounds f ir
           BS.putStr
             $ T.E.decodeUtf8 $ BS.toStrict $ encodePretty graphMessage
    |
290 |       BS.putStr $ T.E.decodeUtf8 $ BS.toStrict $ encodePretty graphMessage
    |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
(or maybe to be more clear, yes it does but the type is wrong)

SamuelMiksits: Okay you can do
```hs
BS.putStr $ BS.toStrict $ encodePretty graphMessage
```
since it is `putStr` and not `putStrLn` it doesn't have a newline on the end

SamuelMiksits: Okay *sigh* I can do
```hs
import qualified Data.ByteString as BS
import qualified Data.ByteString.Char8 as BS8
```
and then
```hs
  BS8.putStrLn $ BS.toStrict $ encodePretty graphMessage
```

klarasm: Maybe something like:
```haskell
import qualified ByteString.Lazy as BSL
-- ...
BSL.putStr $ encodePretty graphMessage
putStrLn ""
```

klarasm: Might be another way to do this which is nicer. I'll look a little.

klarasm: Maybe:
```haskell
diff --git a/src/LSP.hs b/src/LSP.hs
index 41bf48691352..05a0e4f12d0a 100644
--- a/src/LSP.hs
+++ b/src/LSP.hs
@@ -17,15 +17,13 @@ import Control.Monad.IO.Unlift
 import CoreIRToForSyDeIR
 import Data.Aeson ((.=))
 import qualified Data.Aeson as A
-import Data.Aeson.Encode.Pretty
+import qualified Data.Aeson.Encode.Pretty as AP
 import Data.Aeson.KeyMap ((!?))
-import qualified Data.ByteString as BS
+import qualified Data.ByteString.Lazy as BSL
 import qualified Data.List.NonEmpty as NE
 import Data.Proxy
 import qualified Data.Sequence as Seq
 import qualified Data.Text as T
-import qualified Data.Text.Encoding as T.E
-import qualified Data.Text.IO as T.IO
 import ForSyDeIR
 import Language.LSP.Protocol.Message
 import Language.LSP.Server
@@ -287,7 +285,7 @@ run (Arguments (Host ip) (TCP p) i_f) =
       (core, dflags) <- withRunInIO (\_u -> compileToCore f)
       let ir = translateCoreProgram dflags core
       let graphMessage = requestBounds f ir
-      T.IO.putStrLn $ T.E.decodeUtf8 $ BS.toStrict $ encodePretty graphMessage
+      BSL.putStr $ AP.encodePretty graphMessage <> "\n"
 
 -- | Listen on host and port, as well as accept and fork off connections
 runTCPServer :: Maybe HostName -> ServiceName -> (Socket -> IO a1) -> IO a2
```

SamuelMiksits: Okay there is Data.ByteString.Lazy.Char8 which has putStrLn, so I think I am going with
```hs
import qualified Data.ByteString.Lazy.Char8 as BSL8

-- ...

  BSL8.putStrLn $ AP.encodePretty graphMessage

SamuelMiksits: > Okay there is Data.ByteString.Lazy.Char8 which has putStrLn, so I think I am going with
> 
> ```haskell
> import qualified Data.ByteString.Lazy.Char8 as BSL8
> 
> -- ...
> 
>   BSL8.putStrLn $ AP.encodePretty graphMessage
> ```

I implented this now, can I merge?

klarasm: Go ahead :)

PR #259, opened 2025-11-18
Improve KProperties and graphical form

klarasm:
From
SDF_example_006(2)

To
SDF_example_006(3)

From
SDF_example_002(2)

To
SDF_example_002(3)

### DIFF
--- a/src/LSP.hs
+++ b/src/LSP.hs
@@ -68,9 +68,9 @@ forSyDeIRToGraph file (IRSystem (inputs, outputs) actors signals _) = graph
           (createPortWithRate)
           (Just name)
           [KEllipse [KBackgroundColor 160 160 240]]
-          [ (NodeLabelsPlacement, [1, 4, 6]),
-            (NodeSizeConstraints, [3]),
-            (NodeSizeMinimum, [64, 64])
+          [ (NodeLabelsPlacement [1, 4, 6]),
+            (NodeSizeConstraints [3]),
+            (NodeSizeMinimum [64, 64])
           ]
       (IRDelay name d _) ->
         createNode'
@@ -78,9 +78,9 @@ forSyDeIRToGraph file (IRSystem (inputs, outputs) actors signals _) = graph
           (createPortWithoutRate)
           Nothing
           [KEllipse [KBackgroundColor 0 0 0]]
-          [ (NodeLabelsPlacement, [1, 4, 6]),
-            (NodeSizeConstraints, [3]),
-            (NodeSizeMinimum, [12, 12])
+          [ (NodeLabelsPlacement [1, 4, 6]),
+            (NodeSizeConstraints [3]),
+            (NodeSizeMinimum [12, 12])
           ]
     -- \| Find all signals which the process is the source of
     findSourceSignals signals proc =
--- a/src/LSP.hs
+++ b/src/LSP.hs
@@ -83,14 +83,14 @@ forSyDeIRToGraph file (IRSystem (inputs, outputs) actors signals _) = graph
             (NodeSizeMinimum [12, 12])
           ]
     -- \| Find all signals which the process is the source of
-    findSourceSignals signals proc =
+    findOutputSignals signals proc =
       foldr f [] signals
       where
         f s acc =
           let IRSignal n (p, rate) _ = s
            in if p == proc then (n, rate) : acc else acc
     -- \| Find all signals which the process is the target of
-    findTargetSignals signals proc =
+    findInputSignals signals proc =
       foldr f [] signals
       where
         f s acc =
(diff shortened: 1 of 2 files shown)


src/SKGraphSchema.hs
```diff
@@ -68,15 +71,21 @@ kArrow =
...
   show EdgeType = "org.eclipse.elk.edge.type"
   show JunctionPoints = "org.eclipse.elk.junctionPoints"
+  show LayerConstraint = "org.eclipse.elk.layered.layering.layerConstraint"
+  show PortBorderOffset = "org.eclipse.elk.port.borderOffset"
 
 -- We might want to change this later if we add the symbolic enum names
-type KProperties = [(KProperty, [Int])]
+type KProperties = [KProperty]
+
+data KProperty
```
SamuelMiksits: Ah okay, I thought `KPropertySingle` was that you had only 1 property inside `KProperties`, but I understand now it means you have only a single item (`Int`) for the property. But thats fine. I do wonder though why do we have
```hs
data KProperty
  = KPropertyList KPropertyName [Int]
  | KPropertySingle KPropertyName Int
```
and not something like
```hs
data KProperty
  = NodeLabelsPlacement [Int]
-- ...
  | PortBorderOffset Float
```
I suppose the `instance Show KProperty where` would be a bit more difficult but couldn't you do something like say:
```hs
instance show KProperty where
-- ...
  show PortBorderOffset d = "org.eclipse.elk.port.borderOffset: " ++ d
```
(obviously this needs to be done for all of them since otherwise you will get something like "oh the instances don't have the same amount of arguments). Maybe in this case it is fine since it seems to accept -7 as a `double` even though it is missing the decimal point, but what about `bool`? Could otherwise be a fine compromise if it turns out we don't need other properties).

But it also seems a bit weird that it isn't a ToJSON instance? I feel like I am missing something

SamuelMiksits (commented):
Two other points, I can't find a good place to put them in the code:

- It looks like the label text inside an actor is not centered, and now I can't unsee it :)
- Is it possible to have say "⯈" instead of "⧓" specifically on connections to some "s_out"?

But both of these points are about the arrowhead hack, which we probably shouldn't waste too much more time on.

src/SKGraphSchema.hs
klarasm: I also think it would be better to have a ToJSON instance, but couldn't really figure out how to do that yet. I separated them for it to be easier (less case matches) but, yeah, having this inside the property itself would probably be better.

SamuelMiksits: Wouldn't it be similar to how I created coordinates? From KPolygon I have:
```hs
KPolygon styles p ->
      A.object
        [ "type" .= T.pack "KPolygonImpl",
          "children" .= (Seq.empty :: Seq.Seq A.Object),
          "actions" .= (Seq.empty :: Seq.Seq A.Object),
          "styles" .= styles,
          "points" .= (Seq.fromList (buildCoords p)),
          "properties"
            .= A.object
              [ "klighd.lsp.rendering.id" .= T.pack "$R0"
              ]
        ]
```
where `"points"` is of interest, that uses
```hs
          "points" .= (Seq.fromList (buildCoords p)),
```
where then I have
```hs
buildCoords :: [(KPlacementData, KPlacementData)] -> [A.Value]
buildCoords placements = case placements of
  (p1, p2) : xs ->
    A.object
      [ "x" .= p1,
        "y" .= p2
      ]
      : buildCoords xs
  [] -> []
```
The `p1` and `p2` themselves are `KPlacementData` which themselves have

```hs
instance A.ToJSON KPlacementData where
  toJSON placement = case placement of
    KTopPosition absolute relative -> simple "KTopPositionImpl" absolute relative
    KLeftPosition absolute relative -> simple "KLeftPositionImpl" absolute relative
    KRightPosition absolute relative -> simple "KRightPositionImpl" absolute relative
    KBottomPosition absolute relative -> simple "KBottomPositionImpl" absolute relative
    where
      simple name absolute relative =
        A.object
          [ "type" .= T.pack name,
            "absolute" .= absolute,
            "relative" .= relative
          ]

```

klarasm: The thing is that the properties is a dictionary, not a list I believe so it's a little different. But something similar is probably possible.

SamuelMiksits: Oh I had missed that. There is [`Map.fromList`](https://hackage.haskell.org/package/aeson-2.2.3.0/docs/Data-Aeson.html#g:17) instead of `Seq.fromList` but I haven't tried it (I guess this is what `M.fromList` is in your code)

sthaeron:
I think the black rhombus as an arrow was a fine solution. The blue bow makes the diagrams kind of hard to read. I would suggest keeping the black rhombus and changing the actors from circles to rounded squares to improve readability.

klarasm:
Okay. I dropped the bow symbol for now, it looks like:
SDF_example_006

and
SDF_example_002(2)

src/SKGraphSchema.hs
klarasm: I tried to make this a bit less horrible. Unfortunately I can't add a `ToJSON` instance to `KProperties` directly as it complains about overlapping instances.

src/LSP.hs
```diff
@@ -100,10 +100,10 @@ forSyDeIRToGraph file (IRSystem (inputs, outputs) actors signals _) = graph
     createNode' name createPort l r p = node
       where
         nid = T.pack ("$root$N" ++ name)
-        insignals = findSourceSignals signals name
-        outsignals = findTargetSignals signals name
-        inports = map (createPort nid []) insignals
-        outports = map (createPort nid (maybe [] (\_l -> [KText "◆" []]) l)) outsignals
+        insignals = findInputSignals signals name
+        outsignals = findOutputSignals signals name
+        inports = map (createPort nid (maybe [] (\_l -> [KText "◆" []]) l) []) insignals
```
SamuelMiksits, line 105: this diff is making me laugh so hard XD

SamuelMiksits (approved):
Okay, I think this is at a point where it might not be as clean as you maybe want (which might not be possible). But it is a lot more flexible if you need to add a new `KProperties` in the future and it is good enough

src/LSP.hs
SamuelMiksits: ok nvm I misread it, it was fine

Issue #199, opened 2025-11-12
Adding location info to ForSyDe IR


sthaeron:
`new chaos`

### What to change
GHC Core has a type that stores file name, line number, and column for Core Expressions in the original Haskell code. This is done through a `Tick` and is enabled when the `DynFlags` have `debugLevel = 1` and `backend = interpreterBackend`. We need to add a feature that interprets these ticks and adds file name, line number, and column to each ForSyDe IR constructor, signal, and function.

### How to change
Introduce a new data type to ForSyDe IR which stores a file name, starting line number, starting column, ending line number, and ending column (Same info stored in a GHC Core `Tick`). Add this new location type as a optional part of ForSyDe IR constructors, signals, and functions.  Add pattern matching options to core translation when identifying process constructors, functions, system inputs, applying process constructors, and system outputs.

### Why to change
Not high priority but it would allow the visualiser to take advantage of this new location information to make interactive diagrams that when clicked highlights and gotos a specific line and column in the haskell code.

### When to change
When ever we fill the need to upgrade/finalise our current end to end visualiser prototype. Probably after or in parallel to packaging our lsp into a VSCode extenstion. Note the feature will probably only work with the KLighD VSCode exentsion and not the KLighD CLI.

sthaeron:
The following is the core output of the system for SDF example 3 with `debugLevel = 1`, to demonstrate how the `Tick` type is used in GHC Core:
```
NonRec(system =
  Lam(s_in ->
  Tick(src<examples/model/SDF_example_003.hs:(7,1)-(10,17)>: Let(
    Rec({
      (s_1 =
        Case(Var(ds_d19t) of wild_00 {
          Alt(DataAlt((,)): {s_out, s_1} =
            Var(s_1)
          )
        })
      ),
      (s_2 =
        Tick(src<examples/model/SDF_example_003.hs:10:5-17>: Tick(src<examples/model/SDF_example_003.hs:10:11-17>: App(Var(d_1) *
        Tick(src<examples/model/SDF_example_003.hs:10:15-17>: Var(s_1)))))
      ),
      (ds_d19t =
        Tick(src<examples/model/SDF_example_003.hs:9:20-31>: App(App(Var(a_a) *
        Tick(src<examples/model/SDF_example_003.hs:9:24-27>: Var(s_in))) *
        Tick(src<examples/model/SDF_example_003.hs:9:29-31>: Var(s_2))))
      )
    })
   in
    Let(
[...]

PR #260, opened 2025-11-19
Show code related to diagram object

klarasm:
Uses the location info in #265 and should not be merged before it.

This should select the code related to the diagram object when that is clicked (though it will currently open a new window doing that).

### DIFF
--- a/forsyde-devtools.cabal
+++ b/forsyde-devtools.cabal
@@ -30,6 +30,7 @@ common common-options
     ghc-paths ==0.1.0.12,
     hmatrix ==0.20.2,
     lsp ==2.7.0.1,
+    lsp-types ==2.3.0.1,
     network ==3.2.7.0,
     optparse-applicative >=0.18.1.0,
     prettyprinter ==1.7.1,
--- a/src/LSP.hs
+++ b/src/LSP.hs
@@ -216,20 +216,29 @@ handlers f =
         c <- getConfig
         let file = maybe (getFile p) id c
         _ <- setConfig (Just file)
+
+        -- Decode elementSelected
         let sel = getSelected p & map (T.split (\_c -> _c == '$')) & map last & map T.unpack
-        _ <- withRunInIO (\_u -> putStrLn $ show $ sel)
-        sendNotification diagramAcceptMethod setSynthesis
-        sendNotification diagramAcceptMethod (updateOptions file)
+
+        -- Does the client want a diagram?
+        let update = shouldUpdate p
+
+        -- TODO: maybe only recompute on TextDocumentDidSave
         (core, dflags) <- withRunInIO (\_u -> compileToCore file)
         let (forsydeIR, _lookupSignals) = translateCoreProgram dflags core
+
+        -- Get location information on selected object
         let IRSystem _ procs sigs _ = forsydeIR
         let s = map findSignalSpan (map IRString sel) <*> [sigs] & mconcat
         let a = map findProcessSpan (map IRString sel) <*> [procs] & mconcat
         let spans = s ++ a
         _ <- if length spans > 0 then withRunInIO (\_u -> putStrLn $ show spans) else pure ()
 
+        -- Send the diagram if the client wants it
+        if update then sendNotification diagramAcceptMethod setSynthesis else pure ()
+        if update then sendNotification diagramAcceptMethod (updateOptions file) else pure ()
         let graphMessage = requestBounds file forsydeIR
-        sendNotification diagramAcceptMethod graphMessage
+        if update then sendNotification diagramAcceptMethod graphMessage else pure ()
         pure ()
     ]
   where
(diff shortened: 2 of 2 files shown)


SamuelMiksits:
@klarasm What do you want me to do here? I feel a bit lost. I saw that @ramadhanafif posted a tar gzip in Discord with a sample LSP, but I don't know what I should do for it to be considered a "review". 

1. Do I need the "location info"-PR commits? 
2. Should I just run this with the `klighd-linux" CLI client and click something in the browser and have that select code?
3. Should I just inspect the produced JSON? (and what then is the expected output)
4. Do I need to launch the VSCode extension in some way and then have this feature actually work where you have a JS Webview open where you can click a signal and then some code gets highlighted?
5. (I guess I will just run `cabal test` because you changed a lot of things in the test suite. But I am not sure, did you add new tests?)

klarasm:
> @klarasm What do you want me to do here? I feel a bit lost. I saw that @ramadhanafif posted a tar gzip in Discord with a sample LSP, but I don't know what I should do for it to be considered a "review".
> 
>    1. Do I need the "location info"-PR commits?
> 
>    2. Should I just run this with the `klighd-linux" CLI client and click something in the browser and have that select code?
> 
>    3. Should I just inspect the produced JSON? (and what then is the expected output)
> 
>    4. Do I need to launch the VSCode extension in some way and then have this feature actually work where you have a JS Webview open where you can click a signal and then some code gets highlighted?
> 
>    5. (I guess I will just run `cabal test` because you changed a lot of things in the test suite. But I am not sure, did you add new tests?)

Sorry, this is a bit of a mess. This PR is based on #265 (which in turn is based on #266) so you only need to look at this PR.
Testing it in vscode is currently a little painful. You need to:

1. Download and unpack the lsp-sample which @ramadhanafif posted
2. Run `npm install` in the extracted directory (as well as the `client` subdirectory)
3. Run `npm run compile` in the extracted directory
4. Run the lsp separately
5. Launch vscode with `code --extensionDevelopmentPath=$PWD`

Then you can open a model. It seems that it does not always remember the client layout setting, so if there's no diagram, check that.

SamuelMiksits:
> Sorry, this is a bit of a mess. This PR is based on #265 (which in turn is based on #266) so you only need to look at this PR. Testing it in vscode is currently a little painful. You need to:
> 
> 1. Download and unpack the lsp-sample which @ramadhanafif posted
> 2. Run `npm install` in the extracted directory (as well as the `client` subdirectory)
> 3. Run `npm run compile` in the extracted directory
> 4. Run the lsp separately
> 5. Launch vscode with `code --extensionDevelopmentPath=$PWD`
> 
> Then you can open a model. It seems that it does not always remember the client layout setting, so if there's no diagram, check that.

Okay I tried following these steps, but from the #276 branch. I first get the error
```
[Error - 10:01:11 AM] Language Server Example client: couldn't create connection to server.
Launching server using command /home/sam/Documents/IL2232/forsyde-devtools/vscode-ext/server/forsyde-lsp-exe failed. Error: spawn /home/sam/Documents/IL2232/forsyde-devtools/vscode-ext/server/forsyde-lsp-exe ENOENT
```
I copied over the LSP binary in that location, and now it disappears. But both before and after I did that, whenever I try to click "Open in Diagram" on a Haskell file I just get
```
command 'klighd-vscode.diagram.open' not found
```

ramadhanafif:
> Okay I tried following these steps, but from the #276 branch. I first get the error

I think I forgot to mention in the docs, that for the latest one to work, you need to have env variable "dev" to be set to something. So try it again with:

```bash
dev=true code --extensionDevelopmentPath=$PWD
```

without the `dev=true`, it will assume a production environment, and try to launch `forsyde-lsp-exe` executable. 

What we want for dev environment is to connect over socket on a pre-existing `forsyde-lsp-exe`.

ramadhanafif (approved):
Tested via `cabal run forsyde-lsp-exe` and works well.

SamuelMiksits (commented):
Oh dear this touches a lot of things since the whole String to IRVar change causes everything to need "show" before it now. The tests pass and it looks reasonable. 

The only thing that I can point out related to location info and code highlighting is that if I use KGT with KIELER, and have the diagram and code side-by-side, say the code on the left side and the diagram on the right side. Then what happens is that if I click something in the diagram view, then the code will be highlighted on the left side of the editor. What ends up happening here is that if I have the code on the left side and have the diagram on the right side, it will open up the Haskell file again on the right side and take me away from the diagram view when it takes me to the code (so now I have the same Haskell file open on both the left and right side of the screen). I am not sure if this is an issue on the LSP side or on the extension side.

klarasm:
> Oh dear this touches a lot of things since the whole String to IRVar change causes everything to need "show" before it now. The tests pass and it looks reasonable.
> 
> The only thing that I can point out related to location info and code highlighting is that if I use KGT with KIELER, and have the diagram and code side-by-side, say the code on the left side and the diagram on the right side. Then what happens is that if I click something in the diagram view, then the code will be highlighted on the left side of the editor. What ends up happening here is that if I have the code on the left side and have the diagram on the right side, it will open up the Haskell file again on the right side and take me away from the diagram view when it takes me to the code (so now I have the same Haskell file open on both the left and right side of the screen). I am not sure if this is an issue on the LSP side or on the extension side.

Yes, this is an issue. Right now this just uses the LSP window/showDocument message, but KLighD-VSCode might have a better way to do this through their actions hooks.

klarasm:
This should be ready now that #199 is merged (and also easier to review since the unrelated changes are not part of it anymore).

PR #261, opened 2025-11-20
Add shell completions for bash/fish/zsh and add executables to path

### LINKED ISSUE #273: Shell completion for options and arguments for the binaries
Enhance the user experience of the tools by adding so that:
- Options/flags/arguments can be tab-completed
- Add completions that work for different shell types (bash/fish/zsh)

SamuelMiksits:
Yay I created shell completion! This adds the following:

- `forsyde-devtools-exe` and `forsyde-lsp-exe` to `PATH`. This makes it so you can run `forsyde-devtools-exe` directly without the usage of `cabal exec forsyde-devtools-exe --` and having to manually type that out because `cabal exec` does not tab match.
- Shell completion for bash/fish/zsh for options inside those 2 executables.

## How it works

The file `/.completions/generate_completions.sh` is a bash script that generates completion scripts for bash/fish/zsh automatically. I created this since we might want to have this as a pre-commit hook or an automatic build task in the future so that we don't manually have to rerun it.

Then, in the nix flake I use the following

```sh
# Add completion script
source .completions/forsyde-devtools-exe/$0
source .completions/forsyde-lsp-exe/$0
```
where `$0` retrieves the name of the current shell. This is where I am not entirely sure. If I run this in a "normal terminal" I get for instance:

```
$ echo $0
/bin/bash 
```

but if I run this inside `nix develop` I get:

```
$ echo $0
bash 
```

But if this can only be ran inside nix this should probably be fine? Then it should only be `bash` or `zsh` or `fish` without the `/bin/` part in front of it.

Also, sorry, I couldn't help myself. I changed `chmod +x .git/hooks/*` to `chmod u+x .git/hooks/*` and didn't undo it before I commited, I can edit it back if you want.

## Test
Start with 
```
$ nix develop
```
from the root of `forsyde-devtools`. This should add `forsyde-devtools-exe` and `forsyde-lsp-exe` to `PATH`.
Then, try to enter `forsyde-devtools-exe` by typing
```
forsyde-d<TAB>
```
which should automatically shell complete to `forsyde-devtools-exe`. 
Try that this also works with:
```
forsyde-l<TAB>
```
which should automatically shell complete to `forsyde-lsp-exe`.
Then, try to complete some options in the executable such as `--output-forsyde-ir` with
```
forsyde-d<TAB> --output-f<TAB>
```
which should complete to `forsyde-devtools-exe --output-forsyde-ir`.

Also, try that it works for the LSP with:

```
forsyde-l<TAB> --p<TAB>
```
which should complete to `forsyde-lsp-exe --port`.

### DIFF
--- a/.completions/forsyde-devtools-exe/bash
+++ b/.completions/forsyde-devtools-exe/bash
@@ -1,0 +1,14 @@
+_forsyde-devtools-exe()
+{
+    local CMDLINE
+    local IFS=$'\n'
+    CMDLINE=(--bash-completion-index $COMP_CWORD)
+
+    for arg in ${COMP_WORDS[@]}; do
+        CMDLINE=(${CMDLINE[@]} --bash-completion-word $arg)
+    done
+
+    COMPREPLY=( $(/home/sam/Documents/IL2232/forsyde-devtools/dist-newstyle/build/x86_64-linux/ghc-9.10.2/forsyde-devtools-0.0.0.1/x/forsyde-devtools-exe/build/forsyde-devtools-exe/forsyde-devtools-exe "${CMDLINE[@]}") )
+}
+
+complete -o filenames -F _forsyde-devtools-exe forsyde-devtools-exe
--- a/.completions/forsyde-devtools-exe/fish
+++ b/.completions/forsyde-devtools-exe/fish
@@ -1,0 +1,19 @@
+ function _forsyde-devtools-exe
+    set -l cl (commandline --tokenize --current-process)
+    # Hack around fish issue #3934
+    set -l cn (commandline --tokenize --cut-at-cursor --current-process)
+    set -l cn (count $cn)
+    set -l tmpline --bash-completion-enriched --bash-completion-index $cn
+    for arg in $cl
+      set tmpline $tmpline --bash-completion-word $arg
+    end
+    for opt in (/home/sam/Documents/IL2232/forsyde-devtools/dist-newstyle/build/x86_64-linux/ghc-9.10.2/forsyde-devtools-0.0.0.1/x/forsyde-devtools-exe/build/forsyde-devtools-exe/forsyde-devtools-exe $tmpline)
+      if test -d $opt
+        echo -E "$opt/"
+      else
+        echo -E "$opt"
+      end
+    end
+end
+
+complete --no-files --command forsyde-devtools-exe --arguments '(_forsyde-devtools-exe)'
--- a/.completions/forsyde-devtools-exe/zsh
+++ b/.completions/forsyde-devtools-exe/zsh
@@ -1,0 +1,32 @@
+#compdef forsyde-devtools-exe
+
+local request
+local completions
+local word
+local index=$((CURRENT - 1))
+
+request=(--bash-completion-enriched --bash-completion-index $index)
+for arg in ${words[@]}; do
+  request=(${request[@]} --bash-completion-word $arg)
+done
+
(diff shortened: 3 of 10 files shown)


SamuelMiksits:
@sthaeron You use zsh right? So would be good if you could test it for that

SamuelMiksits:
Converted to draft since I was having some issues. When I try to pass a file it can't find it which is kind of a problem
(EDIT: Okay seems like it works, but the issue is that bash completion for arguments doesn't work, since it is not enabled by default. This can be fixed by adding `action`:

```hs
-- Handle file input, always need to define a file
inputFile :: Parser Input
inputFile =
  InputFile
    <$> strArgument
      ( metavar "INPUT"
          <> help "Input filename"
          <> action "file"
      )
```

When I run this in my normal bash it works fine, but when I run in Nix I get
```
bash: line 1: compgen: command not found
```
this is typically a problem when you are in a non-interactive shell so I am stuck troubleshooting that right now)

.completions/forsyde-devtools-exe/bash
```diff
@@ -0,0 +1,14 @@
...
+{
+    local CMDLINE
+    local IFS=$'\n'
+    CMDLINE=(--bash-completion-index $COMP_CWORD)
+
+    for arg in ${COMP_WORDS[@]}; do
+        CMDLINE=(${CMDLINE[@]} --bash-completion-word $arg)
+    done
+
+    COMPREPLY=( $(/home/sam/Documents/IL2232/forsyde-devtools/dist-newstyle/build/x86_64-linux/ghc-9.10.2/forsyde-devtools-0.0.0.1/x/forsyde-devtools-exe/build/forsyde-devtools-exe/forsyde-devtools-exe "${CMDLINE[@]}") )
```
SamuelMiksits, line 11: Hmm, I guess it uses my "PWD" so it puts literally my path in there

ramadhanafif:
This is cool

.completions/forsyde-devtools-exe/bash
klarasm: Ah, yes. After replacing that with my own path it worked for me.

SamuelMiksits: I guess the intention is that you should re-generate this script by running
```
$ ./.completions/generate_completions.sh
```
inside NixOS from the root `forsyde-devtools/`. My general problem is that the [wiki](https://github.com/pcapriotti/optparse-applicative/wiki/Bash-Completion) for `optparse-applicative` says the following:

> For a quick test, you can run something like (for a program called foo on the PATH):
> ```
> $ source <(foo --bash-completion-script `which foo`)
> ```
> Normally, the output of --bash-completion-script should be shipped with the program and copied to the appropriate directory (usually /etc/bash_completion.d/) during installation.

and I don't understand how I can ship this script with our project unless the binary is in like `/usr/bin` or something (since `which` resolves the path to something on my home path). 

I can't really figure out how to get a `PATH` to our binary inside the bash completions that can calculate itself, that we can actually ship. Like do we literally have to put `$(pwd)` inside the completion scripts ourselves? I can put a command that runs the `.completions/generate_completions.sh` inside the Nix shell hook, but the issue is that we don't know if the binary exists yet when you do that (on first time clone of the repo I don't expect it to be built), so the `which` command would probably error out.

SamuelMiksits:
@sthaeron Now I have updated it so that you can autocomplete file paths for arguments, for both the compiler and LSP server. So this means that now you can do:
```
forsyde-devtools-exe --o<TAB>-f<TAB> ./e<TAB>m<TAB>S<TAB>03<TAB> --s<TAB>
```
(I mean you can technically also autocomplete the `forsyde-devtools-exe` binary)
to get to
```
forsyde-devtools-exe --output-forsyde-ir ./examples/model/SDF_example_003.hs --stdout
```
where it was previously not possible to autocomplete the path to the model. So we can leave this for now and revisit it once we have a real shippable binary.

PR #262, opened 2025-11-20
Allow up to 4 global system outputs

klarasm:


sthaeron (approved):
I tested it myself and everything worked as expected. I just added a commit which adds some simple SDF examples with 3 and 4 system outputs, since I had to create them to test the new feature.

PR #263, opened 2025-11-20
get variable name without pretty-printing it

klarasm:
This does not need the dynflags which I think is nice.


klarasm:
Hmm. Unfortunately we don't get unique names with this, which breaks our assumptions in the translation.

PR #265, opened 2025-11-21
Add location information to ForSyDeIR

klarasm:
Add location info to ForSyDeIR. This builds upon #266 which was most of the invasive part.

Also clean up some functions which are just reimplementations of a fold.

### DIFF
--- a/src/CoreIRToForSyDeIR.hs
+++ b/src/CoreIRToForSyDeIR.hs
@@ -296,19 +296,15 @@ createSignalsFromArguments context pcId arguments =
   let inputRates = case (lookup pcId (pcRates context)) of
         Just (rates, _) -> rates
         Nothing -> error ("createSignalsFromArguments - No rates found for process constructor: " ++ show pcId)
-   in aux context (reverse arguments) (reverse inputRates)
+   in foldr aux context $ zip arguments inputRates
   where
-    aux :: TranslationContext -> [IRId] -> [Int] -> TranslationContext
-    aux currentContext currentArguments rates = case (currentArguments, rates) of
-      ([], _) -> currentContext
-      (_, []) -> currentContext
-      (argumentsHead : argumentsTail, ratesHead : ratesTail) ->
-        let (sourceId, sourceRate) = getSourceFromArgument currentContext argumentsHead
-            newSignal = IRSignal argumentsHead (sourceId, sourceRate) (pcId, ratesHead)
-            newSignals = (argumentsHead, newSignal) : (signals currentContext)
-            context1 = currentContext {signals = newSignals}
-            context2 = updateConstructorsInputs context1 pcId argumentsHead
-         in aux context2 argumentsTail ratesTail
+    aux :: (IRId, Int) -> TranslationContext -> TranslationContext
+    aux (argument, rate) currentContext =
+      let (sourceId, sourceRate) = getSourceFromArgument currentContext argument
+          newSignal = IRSignal argument (sourceId, sourceRate) (pcId, rate)
+          newSignals = (argument, newSignal) : (signals currentContext)
+          context1 = currentContext {signals = newSignals}
+       in updateConstructorsInputs context1 pcId argument
 
 -- | Updates the inputs of constructors within a `TranslationContext`. Adds a
 -- signal to the head of a process constructors input signals list.
--- a/src/CoreIRToForSyDeIR.hs
+++ b/src/CoreIRToForSyDeIR.hs
@@ -157,33 +157,29 @@ updateSystemOutput initialContext output =
 updateConstructorsAndSignals :: TranslationContext -> TranslationContext
 updateConstructorsAndSignals initialContext =
   let initialSignals = map snd (signals initialContext)
-      context1 = aux initialContext initialSignals []
-   in context1
+      (context1, sigs) = foldl' aux (initialContext, []) initialSignals
+   in context1 {signals = sigs}
   where
-    aux :: TranslationContext -> [IRSignal] -> [(IRId, IRSignal)] -> TranslationContext
-    aux currentContext currentSignals acc = case currentSignals of
-      [] ->
-        let context1 = currentContext {signals = acc}
-         in context1
-      currentSignal@(IRSignal signalId (sourceId, sourceRate) (targetId, targetRate)) : signalsTail ->
-        let maybebinder = lookup sourceId (binders currentContext)
-         in case maybebinder of
-              Just associatedbinder -> case associatedbinder of
-                PcId pcId ->
-                  -- Signal source was a temporary binder, sourceRate represents
-                  -- the index of the output rather than a rate
-                  let outputRates = case (lookup pcId (pcRates currentContext)) of
-                        Just (_, rates) -> rates
-                        Nothing -> error ("updateConstructorsAndSignals - No rates found for process constructor: " ++ show pcId)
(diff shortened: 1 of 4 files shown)


klarasm:
Hmm. I'm considering whether it might be better to just
```haskell
data IRVar = IRVar Var

instance Show IRVar where
  show (IRVar i) = (occNameString $ nameOccName $ varName i)

instance Eq IRVar where
  (==) (IRVar a) (IRVar b) = varUnique a == varUnique b
```
and then pass IRVars instead of strings. This should also be more efficient for comparison since it's done with the uniques.

We could then get the location from the var directly, and get away without having to change the number of elements in the IR types. We might also get the type information this way.

(marking this as draft again due to #266)

klarasm:
(This is now based on #266)

src/CoreIRToForSyDeIR.hs
```diff
@@ -298,19 +292,15 @@ createSignalsFromArguments context pcId arguments =
   let inputRates = case (lookup pcId (pcRates context)) of
         Just (rates, _) -> rates
         Nothing -> error ("createSignalsFromArguments - No rates found for process constructor: " ++ show pcId)
-   in aux context (reverse arguments) (reverse inputRates)
+   in foldr aux context $ zip arguments inputRates
```
sthaeron, line 295: We started switching to `foldl'` to increase the readability of the code and to ensure that it evaluates the tail-recursive function strictly. Thus, should we be avoiding `foldr` or do you think it does not matter?

klarasm: It should probably not matter. `foldr` does not have the same linear stack space requirements as regular `foldl` (it does not need to build up a thunk as it can lazily just take one element at a time). `foldl'` might still be more efficient as we know we have finite lists, but it shouldn't be a large difference.

klarasm: So my opinion is basically, if `foldr` makes the code more readable (or if we actually need the laziness), then use that. Otherwise always use `foldl'` over `foldl` if possible.

sthaeron:
The original chaos ticket #199, this PR is resolving, mentioned adding a location info type to ForSyDe IR. Meaning I think we should change the ForSyDe IR types to include an optional `IRSpan` such as:
```haskell
data IRConstructor
  = IRDelay IRId [Int] (IRId, IRId) (Maybe IRSpan)
  | IRActor IRId ActorType IRId ([IRId], [IRId]) (Maybe IRSpan)
data IRSignal = IRSignal IRId (IRId, Int) (IRId, Int) (Maybe IRSpan)
data IRFunction = IRFunction IRId (Maybe CoreExpr) (Maybe IRSpan)
```
This would formally make it part of ForSyDe IR and `varToSpan` would rather be a function which is part of the Core IR to ForSyDe IR translation module.

sthaeron:
Rather than using show for the pretty printing, there should be an actually pretty printing function for `IRSpan` which returns a string such as `IRSpan("examples/model/SDF_example_008.hs", 25:1, 25:4)`. This would ensure proper spacing, a clear differentiation between line number and column numbers, and detail how the two are grouped together. It would help make debugging and using the pretty printing easier.

klarasm:
> The original chaos ticket #199, this PR is resolving, mentioned adding a location info type to ForSyDe IR. Meaning I think we should change the ForSyDe IR types to include an optional `IRSpan` such as:
> 
> ```haskell
> data IRConstructor
>   = IRDelay IRId [Int] (IRId, IRId) (Maybe IRSpan)
>   | IRActor IRId ActorType IRId ([IRId], [IRId]) (Maybe IRSpan)
> data IRSignal = IRSignal IRId (IRId, Int) (IRId, Int) (Maybe IRSpan)
> data IRFunction = IRFunction IRId (Maybe CoreExpr) (Maybe IRSpan)
> ```
> 
> This would formally make it part of ForSyDe IR and `varToSpan` would rather be a function which is part of the Core IR to ForSyDe IR translation module.

I kind of disagree on this. If we keep it in the id it's easier to reason about which part of it we have the location for (for e.g. signals which have multiple ids). We can also probably have similar accessors for the type information without needing an additional field in each ForSyDeIR type (i.e. more extensible).

sthaeron:
> I kind of disagree on this. If we keep it in the id it's easier to reason about which part of it we have the location for (for e.g. signals which have multiple ids). We can also probably have similar accessors for the type information without needing an additional field in each ForSyDeIR type (i.e. more extensible).

My original idea is that we would want ForSyDe IR to be as independent as possible from GHC, meaning that it could be portable and translatable without it. Nevertheless, that's no longer the case, since we are using `Var` from GHC is part of our `IRId` and `CoreExpr` in `IRFunction`. So basically, your idea is that we just have functions that give us a type or location information. We should just call these functions with a specific `IRId` when we need the information; no need to save it or make it a part of ForSyDe IR?

klarasm:
> > I kind of disagree on this. If we keep it in the id it's easier to reason about which part of it we have the location for (for e.g. signals which have multiple ids). We can also probably have similar accessors for the type information without needing an additional field in each ForSyDeIR type (i.e. more extensible).
> 
> My original idea is that we would want ForSyDe IR to be as independent as possible from GHC, meaning that it could be portable and translatable without it. Nevertheless, that's no longer the case, since we are using `Var` from GHC is part of our `IRId` and `CoreExpr` in `IRFunction`. So basically, your idea is that we just have functions that give us a type or location information. We should just call these functions with a specific `IRId` when we need the information; no need to save it or make it a part of ForSyDe IR?

Yes, exactly. I also kind of like that we won't need to extend ForSyDeIR just because we wanted to add some additional information which we can already access as part of the GHC Var, which would touch almost our entire codebase. Though we can of course find ourselves in a situation we need to extend ForSyDeIR again anyway.

sthaeron:
When testing, I noticed that the pretty printed `IRSpan` would always have an end column number that was 1 more than what Neovim would say when I had my cursor at the end of the variable name in the original Haskell file. I was testing with SDF example 8. It might be an artefact of how GHC determines the end column number or perhaps a problem with Neovim. Not sure how the discrepancy might affect the visualiser when this is used to select variables in VSCode.

sthaeron (approved):
I had one last comment regarding a potential discrepancy in the end column received from GHC for a variable. Also, it is kind of weird that the pretty printing does not match the actual ForSyDe IR types. Nevertheless, I don't think these remarks should block this PR from being merged. Everything looked good to me and it seems to be working as intended.

klarasm:
> When testing, I noticed that the pretty printed `IRSpan` would always have an end column number that was 1 more than what Neovim would say when I had my cursor at the end of the variable name in the original Haskell file. I was testing with SDF example 8. It might be an artefact of how GHC determines the end column number or perhaps a problem with Neovim. Not sure how the discrepancy might affect the visualiser when this is used to select variables in VSCode.

Hmm. I had this issue with VSCode too where it would select off by one. I'll update this and then merge it.

klarasm:
> > When testing, I noticed that the pretty printed `IRSpan` would always have an end column number that was 1 more than what Neovim would say when I had my cursor at the end of the variable name in the original Haskell file. I was testing with SDF example 8. It might be an artefact of how GHC determines the end column number or perhaps a problem with Neovim. Not sure how the discrepancy might affect the visualiser when this is used to select variables in VSCode.
> 
> Hmm. I had this issue with VSCode too where it would select off by one. I'll update this and then merge it.

Weird. When I look in neovim it's correct according to the current implementation. I guess this is something to look for in the future instead.

PR #266, opened 2025-11-24
Get variable name without pretty-printing it, take 2

klarasm:
This takes a little more holistic approach. Basically the idea is that instead of passing around strings, we wrap the `Var` type from GHC instead, like:

```haskell
data IRId
  = IRVar Var
  | IRString String
  | Empty

instance Show IRId where
  show (IRVar i) = occNameString $ nameOccName $ varName i
  show (IRString s) = s
  show Empty = ""

instance Eq IRId where
  (==) (IRVar a) (IRVar b) = varUnique a == varUnique b
  (==) (IRString a) (IRString b) = a == b
  -- NOTE: This implies Empty is not equal to empty
  (==) Empty _ = False
  (==) _ Empty = False
  (==) (IRString a) b = a == show b
  (==) a (IRString b) = show a == b

prettyIRId :: IRId -> String
prettyIRId = show . show
```
Comparisons are done with the `Unique` type, which should keep the properties we want for the CoreIR to ForSyDeIR transformation, as well as being more efficient (not string comparison).

However, we define our own identifiers in e.g the scheduler, and we sometimes want a placeholder in CoreIR to ForSyDeIR transformation (which uses "" as a placeholder which is kind of horrible), which is why we also have `IRString` and `Empty`.

### DIFF
--- a/src/CoreIRToForSyDeIR.hs
+++ b/src/CoreIRToForSyDeIR.hs
@@ -5,7 +5,6 @@ import Data.List (elemIndex)
 import ForSyDeIR
 import GHC hiding (targetId)
 import GHC.Core
-import GHC.Driver.Ppr
 import GHC.Types.Literal
 
 -- | The `TranslationContext` is a data type which is used to pass around
@@ -14,14 +13,14 @@ import GHC.Types.Literal
 -- and functions whilst they are being built during the translation.
 data TranslationContext = TranslationContext
   { flags :: DynFlags, -- Stores `DynFlags` for safely obtaining strings
-    constructors :: [(String, IRConstructor)], -- Associated list of pcIds and IRConstructors
-    signals :: [(String, IRSignal)], -- Associated list of signalIds and IRSignals
-    functions :: [(String, IRFunction)], -- Associated list of functionIds and IRFunctions
-    systemInputs :: [String], -- List of global inputs to net list
-    systemOutputs :: [String], -- List of global outputs to net list
+    constructors :: [(IRId, IRConstructor)], -- Associated list of pcIds and IRConstructors
+    signals :: [(IRId, IRSignal)], -- Associated list of signalIds and IRSignals
+    functions :: [(IRId, IRFunction)], -- Associated list of functionIds and IRFunctions
+    systemInputs :: [IRId], -- List of global inputs to net list
+    systemOutputs :: [IRId], -- List of global outputs to net list
     nameCounter :: Int, -- Counter used for naming signals
-    pcRates :: [(String, ([Int], [Int]))], -- Associated list of pcIds and input and ouput rates
-    binders :: [(String, Binder)] -- Associated list of binderIds and Binders
+    pcRates :: [(IRId, ([Int], [Int]))], -- Associated list of pcIds and input and ouput rates
+    binders :: [(IRId, Binder)] -- Associated list of binderIds and Binders
   }
 
 -- | `Binder` is a data type used to represent what a `CoreBndr` is associated
@@ -31,8 +30,8 @@ data TranslationContext = TranslationContext
 -- associated with a multi-output process constructor and the index identifies
 -- a specific output.
 data Binder
-  = PcId String
-  | Binding String Int
+  = PcId IRId
+  | Binding IRId Int
 
 initialTranslationContext :: DynFlags -> TranslationContext
 initialTranslationContext dflags =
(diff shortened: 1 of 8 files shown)


klarasm:
I opted to not pass along `IRVar` to the procedural IR since we are unlikely to actually need the additional information there.

klarasm:
(moved the `IRVar` definition to `ForSyDeIR.hs` as requested by @sthaeron).

sthaeron:
I would suggest changing `IRVar` to `IRId` since the string that this type is replacing was referred to as an id (identifier) in comments. Also, when the type is being used, the variable name is usually some name and the suffix Id, such as `actorId` or `signalId`.
```haskell
data IRId
  = IRVar Var
  | IRString String
  | Empty
```

sthaeron:
For the pretty printing of `IRVar` type, I found the use of `(show . show)` and `show $ show` quite confusing. I get now that the first show gets a string and the second show shows said string. Nevertheless, I would suggest abstracting it out into a separate function, such as:
```haskell
prettyIRVar :: IRVar -> String
prettyIRVar irVar = show $ show irVar
```
Just for simplicity and then the show would be reserved for actually obtaining the string for comparisons for when translating to Procedural IR.

sthaeron (commented):
Everything worked when I tested it. It's a great idea to reuse the GHC Var type for id comparisons. I had two suggestions regarding the new ForSyde IR type name and pretty printing.

klarasm:
I guess we can merge this separately if it's easier to review that way.

src/CoreIRToForSyDeIR.hs
```diff
@@ -371,46 +370,46 @@ getSourceFromArgument context argument =
 translateCoreExpr :: TranslationContext -> CoreBndr -> CoreExpr -> TranslationContext
 translateCoreExpr context binder expr = case expr of
   Lam _ (App (App (App (Var i) _) _) _)
-    | showPpr (flags context) i == "delaySDF" -> createDelaySDF context binder expr
+    | show (IRVar i) == "delaySDF" -> createDelaySDF context binder expr
```
sthaeron: Is there any difference between this comparison and doing `IRVar i == IRString "delaySDF"`? Should it be changed?

src/ForSyDeIRToProceduralIR.hs
```diff
@@ -239,18 +239,20 @@ translateIRFunctionToGlobals currentContext (IRFunction functionId maybeFunction
...
-        initFunctionGlobal = GFuncDeclare (Just Static) TVoid (binder) [(TPointer TInt, "input_1"), (TPointer TInt, "input_2"), (TPointer TInt, "output_1"), (TPointer TInt, "output_2")]
-        functionGlobal = GFuncDef (Just Static) TVoid (binder) [(TPointer TInt, "input_1"), (TPointer TInt, "input_2"), (TPointer TInt, "output_1"), (TPointer TInt, "output_2")] functionScopeStmt
-     in (initFunctionGlobal, functionGlobal)
-  ("add", Lam _ (Lam _ e)) ->
-    let functionScopeStmt = translateCoreExprToStatement context e
-        initFunctionGlobal = GFuncDeclare (Just Static) TVoid (binder) [(TPointer TInt, "input_1"), (TPointer TInt, "input_2"), (TPointer TInt, "output")]
-        functionGlobal = GFuncDef (Just Static) TVoid (binder) [(TPointer TInt, "input_1"), (TPointer TInt, "input_2"), (TPointer TInt, "output")] functionScopeStmt
-     in (initFunctionGlobal, functionGlobal)
+  (b, Lam _ (Lam _ e))
+    | b == IRString "accumulate" ->
```
sthaeron, line 245: I'm asking since that's what you did here

src/CoreIRToForSyDeIR.hs
klarasm: Hmm. That might be a good idea. I think I didn't do it here since `i` is a regular `var` here, but this approach might be cleaner. I'll change it.

sthaeron:
> merge this separately

That would be great! It'll make reviewing your other PRs much simpler, as the git diff from this PR wouldn't be there any more. It was quite difficult to find what was different in #265.

klarasm:
Btw, when rebasing one can do `git rebase --exec 'cabal build && cabal test' ...` which will check each commit separately and make it more likely to be bisectable in the future. If it fails one can fix that commit and do an amend directly in the rebase.

sthaeron (approved):
It all looks good to me! Could you merge this into dev and rebase #265 based on the new dev?

PR #282, opened 2025-12-01
Argument parsing: add TARGET option for C code

SamuelMiksits:
## Description

Adds the argument option `TARGET` to compiler
- "t" for short option, i.e can do `forsyde-devtools-exe -t PC`
- "target" for long option, i,e can do `forsyde-devtools-ex --target PC` or `forsyde-devtools-ex --target=PC`

Have also added to help text so that it now says:

```
Available options:
  ...
  -t,--target TARGET       Target platform for C code. (PC default, PC and PICO2
                           supported)
  -h,--help                Show this help text
```

## How to run

Check that the tests pass. I added a newline at the end of the file for the `example/test/simple.c`. I also modified so that `translateProgram` now takes platform as argument, so I just inserted `PC` as platform for the test. But since the test is already running without "include" I made it so that this target is omitted when running without include, this is fine right?

Otherwise, try running

```
cabal exec forsyde-devtools-exe -- ./examples/model/SDF_example_008.hs --stdout --output-c --target=PC
```

and check that it has
```
#define PLATFORM PC
typedef int token;
#include "include/common.h"
#include <stdio.h>
```
at the top and that

```
cabal exec forsyde-devtools-exe -- ./examples/model/SDF_example_008.hs --stdout --output-c --target=PICO2
```
has 
```
#define PLATFORM PICO2
typedef int token;
#include "include/common.h"
#include <stdio.h>
```
at the top

## Other comments

I think this way of implementing target raises the following question, why is `output` not implemented in the same way? I.e:
```
--output={Core,ForSyDeIR,ForSyDeIRJSON,ProceduralIR,C}
```
instead of
```
  --output-c
  --output-core
  --output-forsyde-ir
  --output-forsyde-ir-json
  --output-procedural-ir
```
(or possibly have it the other way around, so target is also implemented like that). I currently have it so if you call with an unsupported platform it outputs the following:
```
bash-5.3$ cabal exec forsyde-devtools-exe -- ./examples/model/SDF_example_008.hs --stdout --output-c --target=PC2
Configuration is affected by the following files:
- cabal.project
forsyde-devtools-exe: Unsupported target: PC2
CallStack (from HasCallStack):
  error, called at src/ArgumentsMain.hs:180:12 in forsyde-devtools-0.0.0.1-inplace-forsyde-devtools-exe:ArgumentsMain
HasCallStack backtrace:
[...]

PR #283, opened 2025-12-01
Argument parsing: add output schedule option to compiler

SamuelMiksits:
## Description

This is based on #282 so should probably merge that one first. Adds `--output-schedule` argument option to compiler. This runs the `computeScheduleAndBuffersPrint` function from the scheduler and prints the output. If ran without the `-o` flag it will default to `main.sched` as the filename, and if ran with `--stdout` it will print the output to the terminal.

## How to run

In general just check that the tests run as intended with
```
cabal test
```

Additionally, I changed the function `computeScheduleAndBuffersPrint` so that it adds a newline at the end of the file, so make sure that if you run for instance:

```
cabal exec forsyde-devtools-exe -- ./examples/model/SDF_example_003.hs --stdout --output-schedule
```

that it actually adds a newline at the end of the file

```
bash-5.3$ cabal exec forsyde-devtools-exe -- ./examples/model/SDF_example_003.hs --stdout --output-schedule
Configuration is affected by the following files:
- cabal.project
Topology Matrix (Edges × Actors):
Edge\Actoa_a  
--------------
s_1_s_2  0.0

...

All buffer sizes (I/O + internal):
  s_in: 1
  s_out: 1
  s_1_s_2: 1
bash-5.3$ 
```

### DIFF
--- a/forsyde-devtools.cabal
+++ b/forsyde-devtools.cabal
@@ -42,6 +42,7 @@ library
   import: common-options
   hs-source-dirs: src
   exposed-modules:
+    ArgumentsMain
     CoreIR
     CoreIRToForSyDeIR
     ForSyDeIR
--- a/src/ArgumentsMain.hs
+++ b/src/ArgumentsMain.hs
@@ -31,12 +31,18 @@ data OutputFormat
   | OutputForSyDeIRJSON
   | OutputProceduralIR
 
+data Target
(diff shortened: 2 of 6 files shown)


sthaeron (approved):
Tests work, compiler gets C code, and it all looks good to me!

PR #286, opened 2025-12-02
Add script to test pipeline and c output

Himawari722:


.github/workflows/build.yml
```diff
@@ -37,3 +37,6 @@ jobs:
             nix develop --command bash -c "cabal build"
             nix develop --command bash -c "cabal test"
           fi
+      - name: Run integrated pipeline test
+        run: |
+          bash test_pipeline.sh
```
klarasm, line 42: No final newline.

test_pipeline.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+set -e
+
+echo "Compiling SDF_example_008.hs into C"
+nix run .#forsyde-compiler -- examples/model/SDF_example_008.hs -o ./examples/implementation/platform_independent/test_008.c --output-c
+echo "Compiling C code into executable"
+gcc examples/implementation/platform_independent/test_008.c -o test_008
```
klarasm, line 7: It might be nice to eventually just loop through all the files in the test/model directory, but this should be fine for now since we currently only support this one.

test_pipeline.sh
```diff
@@ -0,0 +1,31 @@
...
+
+expected_output=$(cat <<EOF
```
klarasm, line 9: Here it would be nice to get the output from the haskell model directly but not sure how we would automate that (we would need to write some stub for passing arbitrary list as well as probably adding a cabal project for the test/model directory or something). I think this is fine as it is for now, however.

test_pipeline.sh
```diff
@@ -0,0 +1,31 @@
...
+2
+6
+12
+20
+EOF
+)
+echo "Running..."
+actual_output=$(printf "1 1\n2 2\n3 3\n4 4\n" | ./test_008)
+
+if [ "$expected_output" != "$actual_output" ]; then
```
klarasm, line 19: Maybe we could put this into a function which we call with maybe test name, expected output, and actual output. But that's probably for later once we support more of the models.

klarasm (approved):
I think this looks good. It would probably be good to squash these two commits together.

PR #287, opened 2025-12-02
Properly highlight code from selected diagram

klarasm:
This should now highlight the source code without opening a new window when clicking on a diagram object.

Also remove the unliftio-core dependency.

### DIFF
--- a/forsyde-devtools.cabal
+++ b/forsyde-devtools.cabal
@@ -37,7 +37,6 @@ common common-options
     process ==1.6.25.0,
     syb >=0.7.2.4 && <=0.7.3,
     text ==2.1.2,
-    unliftio-core ==0.2.1.0,
 
 library
   import: common-options
--- a/src/LSP.hs
+++ b/src/LSP.hs
@@ -40,6 +40,9 @@ import Utilities
 diagramAcceptMethod :: SMethod (Method_CustomMethod "diagram/accept")
 diagramAcceptMethod = (SMethod_CustomMethod (Proxy @"diagram/accept"))
 
+diagramOpenInTextEditor :: SMethod (Method_CustomMethod "diagram/openInTextEditor")
+diagramOpenInTextEditor = (SMethod_CustomMethod (Proxy @"diagram/openInTextEditor"))
+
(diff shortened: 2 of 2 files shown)


SamuelMiksits (approved):
Yeah this looks good!

Some other things that I am thinking about:

- In `/docs` we have `vscode.md` and `visualisation.md`, but I feel like neither do a good job of explaining how to run what we have now, and since there is so much stuff in `visualisation.md` about vscode I am wondering if they should even be separate files. I had to go back to #260 again to figure out how to run the VSCode extension and I feel like this is a bit of a problem. Potential action is to create a documentation issue about how to run the vscode extension @ramadhanafif 

- What do we actually want to highlight for certain cases? Okay sure, if you click on an actor it takes you to do process constructor definition for that actor. So for instance if you click on `a_a` for `SDF_example_002.hs` or whatever then it highlights the `a_a` in the line
```hs
a_a s_1 = actor11SDF 2 1 f_1 s_1
```
But if you click on a token rate it takes you to its corresponding signal in the netlist (which clicking the signal label also does). Example is that if you click on the `2` that is on the input side of `a_a` it takes you to `s_in_1` in 
```hs
system s_in_1 s_in_2 = s_out
```
Should clicking the token rate actually highlight that or should it highlight the token rate? I think that can be fine, and depending on what information we have in the IR it may or may not be practical to do that, and is it even desirable to highlight that? Maybe it useful for someone who is porting a visual model to code to be able to click on the token rate that is wrong and be taken to the thing that they need to change to make it correct. I don't think we need to write some doc/spec that defines what clicking on something should take you to, but thought that it could be a good idea to just think about, or point out in case you are intending clicking on token rates to actually take you somewhere else. (dear markdown indenting)

- The only other feature I feel is missing for a reasonable workflow is to have it so that when you are editing a model and you press "save", it would be nice to have the compiler re-run. I think KGraph KGT might even be "realtime" but I think having recompile on save might be better. You can currently press refresh diagram and have it update. Maybe create an issue about this?

klarasm:
I agree with all your points, created issues for each of them.

PR #293, opened 2025-12-03
Add predefined input support to compiler

### LINKED ISSUE #294: Add compiler support for predefined input sequences
Currently the compiler only supports reading inputs from `scanf`, but in the real world someone might want to have predefined data to cycle through

SamuelMiksits:
## Description

Adds support for predefined data from end to end in compiler (from argument parsing, docs, all the way to C code generation).

## How to test

Run `cabal test`, since I have had to change test spec ... again

Since we currently only have the compiler "working" for `examples/model/SDF_example_008.hs` you can run the following tests:

```
cabal exec forsyde-compiler-exe -- ./examples/model/SDF_example_008.hs --output-c --stdout --io=predefined
```

```C
#define PLATFORM PC
typedef int token;
#include "include/common.h"
#include <stdio.h>
#include "input.h"

static void add(int *input_1, int *input_2, int *output);

static void accumulate(int *input_1, int *input_2, int *output_1, int *output_2);

int main() {
    init();
    int status;
    int iter_current = -1;
    buffer_nonblocking *s_in_y = buffer_nonblocking_new(1);
    buffer_nonblocking *s_in_x = buffer_nonblocking_new(1);
    buffer_nonblocking *s_out = buffer_nonblocking_new(1);
    buffer_nonblocking *s_1 = buffer_nonblocking_new(1);
    buffer_nonblocking *s_2_s_2_delayed = buffer_nonblocking_new(1);
    int output_s_out;
    write_token(s_2_s_2_delayed, 0);
    while (1) {
        iter_current = iter_current + 1;
        if (iter_current == iter_max) {
            iter_current = 0;
        }
        for (int i = 0; i < 1; i++) {
            write_token(s_in_x, input_s_in_x[iter_current * 1 + i]);
        }
        for (int i = 0; i < 1; i++) {
            write_token(s_in_y, input_s_in_y[iter_current * 1 + i]);
        }
        actor21SDF(1, 1, 1, s_in_x, s_in_y, s_1, add);
        actor22SDF(1, 1, 1, 1, s_1, s_2_s_2_delayed, s_out, s_2_s_2_delayed, accumulate);
        for (int i = 0; i < 1; i++) {
            read_token(s_out, &output_s_out);
            printf("%d", output_s_out);
        }
        printf("\n");
    }
    buffer_nonblocking_free(s_in_y);
    buffer_nonblocking_free(s_in_x);
    buffer_nonblocking_free(s_out);
    buffer_nonblocking_free(s_1);
    buffer_nonblocking_free(s_2_s_2_delayed);
    return 0;
}

static void add(int *input_1, int *input_2, int *output) {
    output[0] = input_1[0] + input_2[0];
}

static void accumulate(int *input_1, int *input_2, int *output_1, int *output_2) {
    output_1[0] = input_1[0] + input_2[0];
    output_2[0] = input_1[0] + input_2[0];
}
```

the important parts are that it has:

```C
#include "input.h"
```
and
```C
int output_s_out;
```
but no inputs are defined. Then check the while loop, it should have
```C
iter_current = iter_current + 1;
if (iter_current == iter_max) {
    iter_current = 0;
}
```
and then finally instead of what we normally have for input processing it should have
```C
for (int i = 0; i < 1; i++) {
    write_token(s_in_x, input_s_in_x[iter_current * 1 + i]);
}
for (int i = 0; i < 1; i++) {
    write_token(s_in_y, input_s_in_y[iter_current * 1 + i]);
}
```
An example file of `input.h` is the following:
```C
int iter_max = 4;

token input_s_in_x[4] = {1,2,3,4};
token input_s_in_y[4] = {1,2,3,4};

```

and it should produce the following output if you run it:

```
2
6
12
20
```

Since it will run "indefinitely" you have to insert some sort of escape hooks, for example the following:

```C
int runs = 1;    
int run_current = 0;
```

before the `while` loop, then use 
```C
iter_current = iter_current + 1;
if (iter_current == iter_max) {
    iter_current = 0;
    run_current += 1;
}

if (run_current == runs) {
    break;
}
```
Can also check that without `--io` or `--io=scanf` that it runs normally.

## Other comments

I don't think we currently have a place where it is documented on how to run it in this way (i.e. how to format the input file for `input.h`). I mean we do have a developer guide on how to implement C code generation when a user has started the program that way, but not a "user facing" guide if that makes sense. How do you want to go about implementing that? I mention that the option exists in `--help`:

```
  --io IO                  Source of input tokens for the SDF models in the C
                           code. (scanf default, scanf and predefined supported)
```


SamuelMiksits:
Can also be good (if you find any) to point out comments on the top of functions that I have edited in `src/ForSyDeIRToProceduralIR.hs` that might say "function takes x arguments" and now I have sent more so they are wrong or they say something that is specific to `scanf` but now we have `scanf`/`predefined`

klarasm:
Is there a particular reason we update `iter_current` at the top of the loop? It feels weird to me to initialize it to `-1` and then increment it immediately. For me it would feel more natural to initialize it to `0` and increment at the end of the loop.

Clangd complains about variables in a header file, but it might be specific to C++ (One Definition Rule) so we can probably ignore it, and it's not like this header will be included in several files which while be linked together in compilation anyway.

Related to that, as the size of the array is statically known, we could get the iteration count with `sizeof(array)/sizeof(*array)/tokenRate`, but I'm not sure if that would be a good idea here.

SamuelMiksits:
> Is there a particular reason we update `iter_current` at the top of the loop? It feels weird to me to initialize it to `-1` and then increment it immediately. For me it would feel more natural to initialize it to `0` and increment at the end of the loop.

Thought it would be easier? Idk if it really matters that much but I can go back and change it if it matters that much. But I would have to take a look at how this needs to be implemented in the procedural IR and if it causes any issues.
 
> Clangd complains about variables in a header file, but it might be specific to C++ (One Definition Rule) so we can probably ignore it, and it's not like this header will be included in several files which while be linked together in compilation anyway.

I have never used clangd before, so don't know how to reproduce it and what warning flags (or if it also runs clang-tidy with other linting/formatting rules) it uses. But I mean if I run `-Wall` it complains about a lot of things (like unused functions). Actually it did complain that `int status` was unused to I should probably remove that for `predefined` mode

> Related to that, as the size of the array is statically known, we could get the iteration count with `sizeof(array)/sizeof(*array)/tokenRate`, but I'm not sure if that would be a good idea here.

Grea idea, I don't know how I didn't think of it since I have been using exactly this feature in other projects. In that project we used `static_assert` to ensure that all static variables (arrays) were the same length before compiles. You could try to calculate the shortest amount of iterations that has a "defined behavior". For example if you have `input_a[8]` and `input_b[8]`) but then `a` consumes 1 token and `b` consumes 2 tokens (per period) then from the standpoint of `a` you have 8 iterations and from `b` 4. You could then interpret that as:

- The amount of iterations do not match, therefore it fails with a static assert
- The amount of iterations do not match, therefore use cursed template metaprogramming to figure out which one has the least amount of iterations and use that one as the actual period.

The first one is probably idealistically correct but actually wrong. Because one of the benefits of modelling in Haskell is that you can just feed it an arbitrary amount of tokens on either input and it will just run as long as it can. The issue is we can't actually know how long the input data is "compile time" (our compiler, not gcc/clang or whatever), so pragmas is probably the only way.

So maybe I am talking myself out of doing the second approach, but let's have an example of how the code would have to be written:

```C
token input_a[8] = {1,2,3,4,5,6,7,8};
token input_b[8] = {1,2,3,4,5,6,7,8};
token input_c[9] = {1,2,3,4,5,6,7,8,9};

static const int iterations_a = sizeof(input_a)/sizeof(*input_a)/bufferSizeA;  //bufferSize is not really a "C variable", we have it as  variable inside our compiler
static const int iterations_b = sizeof(input_a)/sizeof(*input_a)/bufferSizeB;
static const int iterations_c = sizeof(input_a)/sizeof(*input_a)/bufferSizeC;

#if (iterations_a <= iterations_b && iterations_a <= iterations_c)
  static const int iterations = iterations_a;
#elif (iterations_b < iterations_a && iterations_b < iterations_c)
  static const int iterations = iterations_b;
#else
  static const int iterations = iterations_c;
#endif
``` 

doesn't this just sound great? ;) (there is probably something wrong in my if expressions but just to show the general idea)

klarasm:
> > Is there a particular reason we update `iter_current` at the top of the loop? It feels weird to me to initialize it to `-1` and then increment it immediately. For me it would feel more natural to initialize it to `0` and increment at the end of the loop.
> 
> Thought it would be easier? Idk if it really matters that much but I can go back and change it if it matters that much. But I would have to take a look at how this needs to be implemented in the procedural IR and if it causes any issues.
> 

If it's not too much effort I would appreciate it.

> > Clangd complains about variables in a header file, but it might be specific to C++ (One Definition Rule) so we can probably ignore it, and it's not like this header will be included in several files which while be linked together in compilation anyway.
> 
> I have never used clangd before, so don't know how to reproduce it and what warning flags (or if it also runs clang-tidy with other linting/formatting rules) it uses. But I mean if I run `-Wall` it complains about a lot of things (like unused functions). Actually it did complain that `int status` was unused to I should probably remove that for `predefined` mode
> 

(forgot to answer this one). I'm using it as the LSP for C in neovim. It should be pretty easy to set it up on VSCode too.

> > Related to that, as the size of the array is statically known, we could get the iteration count with `sizeof(array)/sizeof(*array)/tokenRate`, but I'm not sure if that would be a good idea here.
> 
> Grea idea, I don't know how I didn't think of it since I have been using exactly this feature in other projects. In that project we used `static_assert` to ensure that all static variables (arrays) were the same length before compiles. You could try to calculate the shortest amount of iterations that has a "defined behavior". For example if you have `input_a[8]` and `input_b[8]`) but then `a` consumes 1 token and `b` consumes 2 tokens (per period) then from the standpoint of `a` you have 8 iterations and from `b` 4. You could then interpret that as:
> 
>    * The amount of iterations do not match, therefore it fails with a static assert
> 
>    * The amount of iterations do not match, therefore use cursed template metaprogramming to figure out which one has the least amount of iterations and use that one as the actual period.
> 
> 
> The first one is probably idealistically correct but actually wrong. Because one of the benefits of modelling in Haskell is that you can just feed it an arbitrary amount of tokens on either input and it will just run as long as it can. The issue is we can't actually know how long the input data is "compile time" (our compiler, not gcc/clang or whatever), so pragmas is probably the only way.
> 
> So maybe I am talking myself out of doing the second approach, but let's have an example of how the code would have to be written:
> 
> ```c
> token input_a[8] = {1,2,3,4,5,6,7,8};
> token input_b[8] = {1,2,3,4,5,6,7,8};
> token input_c[9] = {1,2,3,4,5,6,7,8,9};
> 
> static const int iterations_a = sizeof(input_a)/sizeof(*input_a)/bufferSizeA;  //bufferSize is not really a "C variable", we have it as  variable inside our compiler
> static const int iterations_b = sizeof(input_a)/sizeof(*input_a)/bufferSizeB;
> static const int iterations_c = sizeof(input_a)/sizeof(*input_a)/bufferSizeC;
> 
> #if (iterations_a <= iterations_b && iterations_a <= iterations_c)
>   static const int iterations = iterations_a;
> #elif (iterations_b < iterations_a && iterations_b < iterations_c)
>   static const int iterations = iterations_b;
> #else
>   static const int iterations = iterations_c;
> #endif
> ```
> 
> doesn't this just sound great? ;) (there is probably something wrong in my if expressions but just to show the general idea)

This is basically just for testing, right? In that case I would prefer the static_assert approach. We can always change it later if we need to.

src/ForSyDeIRToProceduralIR.hs
```diff
@@ -46,21 +47,31 @@ initialTranslationContext dflags lookupSignalList inputList outputList delayBuff
 -- | Translates a ForSyDe IR `IRSystem` into a Procedural IR `Program`. Requires
 -- the schedule, buffer, and delay buffer lists from the SDF scheduler as an
 -- input. Also requires an associated list of signals for lookups.
-translateIRSystemToProgram :: DynFlags -> [IRId] -> [(IRId, Int)] -> [(IRId, IRId)] -> [(IRId, IRSignal)] -> IRSystem -> Program
-translateIRSystemToProgram dflags scheduleList bufferList delayBufferList lookupSignalList (IRSystem (inputList, outputList) constructors _signalList functionList) =
+translateIRSystemToProgram :: DynFlags -> [IRId] -> [(IRId, Int)] -> [(IRId, IRId)] -> [(IRId, IRSignal)] -> IOType -> IRSystem -> Program
+translateIRSystemToProgram dflags scheduleList bufferList delayBufferList lookupSignalList io (IRSystem (inputList, outputList) constructors _signalList functionList) =
```
klarasm: `io` introduces a shadowing warning, in case we care about that.

src/ForSyDeIRToProceduralIR.hs
```diff
@@ -46,21 +47,31 @@ initialTranslationContext dflags lookupSignalList inputList outputList delayBuff
...
-      whileStmt = SWhile (EInt 1) (SScope scheduledStmts)
-      mainInitStmts = [SExpr (ECall "init" []), SVarDecl TInt "status"] ++ reverse (initBuffers context) ++ reverse (ioTokens context) ++ reverse (initDelay context)
+      scheduledStmtsSpecial = case io of
+        Scanf -> []
+        Predefined ->
+          [ SVarAssign "iter_current" (EBinOp Plus (EVar "iter_current") (EInt 1)),
+            SIf (EBinOp Equal (EVar "iter_current") (EVar "iter_max")) (SVarAssign "iter_current" (EInt 0)) Nothing
+          ]
+
+      whileStmt = SWhile (EInt 1) (SScope (scheduledStmtsSpecial ++ scheduledStmts))
```
klarasm: Would it be enough to just switch the order of the concatenation here to increment `iter_current` at the end of the loop?

SamuelMiksits:
> If it's not too much effort I would appreciate it.

okay I could basically go from
```hs
SScope (scheduledStmtsSpecial ++ scheduledStmts)
```
to
```hs
SScope (scheduledStmts ++ scheduledStmtsSpecial)
```
and then I get
```C
int iter_current = 0;
// ...
while (1) {
        for (int i = 0; i < 1; i++) {
            write_token(s_in_x, input_s_in_x[iter_current * 1 + i]);
        }
        for (int i = 0; i < 1; i++) {
            write_token(s_in_y, input_s_in_y[iter_current * 1 + i]);
        }
        actor21SDF(1, 1, 1, s_in_x, s_in_y, s_1, add);
        actor22SDF(1, 1, 1, 1, s_1, s_2_s_2_delayed, s_out, s_2_s_2_delayed, accumulate);
        for (int i = 0; i < 1; i++) {
            read_token(s_out, &output_s_out);
            printf("%d", output_s_out);
        }
        printf("\n");
        iter_current = iter_current + 1;
        if (iter_current == iter_max) {
            iter_current = 0;
        }
    }
```
is this fine?

klarasm:
> > If it's not too much effort I would appreciate it.
> 
> okay I could basically go from
> 
> ```haskell
> SScope (scheduledStmtsSpecial ++ scheduledStmts)
> ```
> 
> to
> 
> ```haskell
> SScope (scheduledStmts ++ scheduledStmtsSpecial)
> ```
> 
> and then I get
> 
> ```c
> int iter_current = 0;
> // ...
> while (1) {
>         for (int i = 0; i < 1; i++) {
>             write_token(s_in_x, input_s_in_x[iter_current * 1 + i]);
>         }
>         for (int i = 0; i < 1; i++) {
>             write_token(s_in_y, input_s_in_y[iter_current * 1 + i]);
>         }
>         actor21SDF(1, 1, 1, s_in_x, s_in_y, s_1, add);
>         actor22SDF(1, 1, 1, 1, s_1, s_2_s_2_delayed, s_out, s_2_s_2_delayed, accumulate);
>         for (int i = 0; i < 1; i++) {
>             read_token(s_out, &output_s_out);
>             printf("%d", output_s_out);
>         }
>         printf("\n");
>         iter_current = iter_current + 1;
>         if (iter_current == iter_max) {
>             iter_current = 0;
>         }
>     }
> ```
> 
> is this fine?

Lol, didn't see this response before I submitted. This looks good to me.

SamuelMiksits:
> Lol, didn't see this response before I submitted. This looks good to me.

I think we both posted at the same time lol

klarasm:
We could perhaps use `__attribute__((unused))` if we want to silence the extra warnings, but it only appears with `-Wall` so we can probably ignore it.

klarasm:
Btw, we have
```c
        for (int i = 0; i < 1; i++) {
            read_token(s_out, &output_s_out);
            printf("%d", output_s_out);
        }
        printf("\n");
```
So if we have a rate higher than 1 the numbers will be printed without any whitespace in between. This is not new in this PR, however.

SamuelMiksits:
> Btw, we have
> 
> ```c
>         for (int i = 0; i < 1; i++) {
>             read_token(s_out, &output_s_out);
>             printf("%d", output_s_out);
>         }
>         printf("\n");
> ```
> 
> So if we have a rate higher than 1 the numbers will be printed without any whitespace in between. This is not new in this PR, however.

Okay I can just add a one liner patch that fixes that. Since I am petty I checked my documentation and I literally say to do
```C
for(i = 0; i < 2; i++) {
    read_token(s_out, &output);
    printf("%d ", output);
}
```
in there :)

Also I was thinking a bit about it, and fundamentally everything we have in the preprocessor right now (defines/includes) are hardcoded strings that we append to the program, and this is because we haven't implemented any preprocessor expressions in IR/code gen. So if we want to do some checks on array length/iterations I feel like we would need to add a substantial amount of things to the IR before that is possible, or do a grotesque hack so I think we might have to create a separate issue for this in the future instead of handling it on this PR. Or idk maybe it isn't actually that bad. Do we just need something like

```C
static const int iterations_a = sizeof(input_a)/sizeof(*input_a)/bufferSizeA;
static const int iterations_b = sizeof(input_a)/sizeof(*input_a)/bufferSizeB;
static const int iterations_c = sizeof(input_a)/sizeof(*input_a)/bufferSizeC;
static_assert(iterations_a == iterations_b == iterations_c);
```
?

klarasm:
> > Btw, we have
> > ```c
> >         for (int i = 0; i < 1; i++) {
> >             read_token(s_out, &output_s_out);
> >             printf("%d", output_s_out);
> >         }
> >         printf("\n");
> > ```
> > So if we have a rate higher than 1 the numbers will be printed without any whitespace in between. This is not new in this PR, however.
> 
> Okay I can just add a one liner patch that fixes that. Since I am petty I checked my documentation and I literally say to do
> 
> ```c
> for(i = 0; i < 2; i++) {
>     read_token(s_out, &output);
>     printf("%d ", output);
> }
> ```
> 
> in there :)
> 
+1

> Also I was thinking a bit about it, and fundamentally everything we have in the preprocessor right now (defines/includes) are hardcoded strings that we append to the program, and this is because we haven't implemented any preprocessor expressions in IR/code gen. So if we want to do some checks on array length/iterations I feel like we would need to add a substantial amount of things to the IR before that is possible, or do a grotesque hack so I think we might have to create a separate issue for this in the future instead of handling it on this PR. Or idk maybe it isn't actually that bad. Do we just need something like
> 
> ```c
> static const int iterations_a = sizeof(input_a)/sizeof(*input_a)/bufferSizeA;
> static const int iterations_b = sizeof(input_a)/sizeof(*input_a)/bufferSizeB;
> static const int iterations_c = sizeof(input_a)/sizeof(*input_a)/bufferSizeC;
> static_assert(iterations_a == iterations_b == iterations_c);
> ```
> 
> ?

This might be enough. Let me check what that does.

klarasm:
I tried:
```c
#include <stddef.h>

int input_a[10],
    input_b[11],
    input_c[12];

const size_t bufferSizeA = 3, bufferSizeB = 4, bufferSizeC = 5;

static const int iterations_a = sizeof(input_a)/sizeof(*input_a)/bufferSizeA;
static const int iterations_b = sizeof(input_b)/sizeof(*input_b)/bufferSizeB;
static const int iterations_c = sizeof(input_c)/sizeof(*input_c)/bufferSizeC;

_Static_assert(iterations_a == iterations_b == iterations_c, "");
```
This works in clang:
```
test.c:13:16: error: static assertion failed due to requirement 'iterations_a == iterations_b ==
      iterations_c':
   13 | _Static_assert(iterations_a == iterations_b == iterations_c, "");
      |                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
test.c:13:45: note: expression evaluates to '0 == 2'
   13 | _Static_assert(iterations_a == iterations_b == iterations_c, "");
      |                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~
1 error generated.
```

but not in gcc:
```
test.c:13:45: error: expression in static assertion is not constant
   13 | _Static_assert(iterations_a == iterations_b == iterations_c, "");
      |                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~
```

klarasm:
If I do
```c
int input_a[10],
    input_b[11],
    input_c[12];

_Static_assert(sizeof(input_a)/sizeof(*input_a)/3
		== sizeof(input_b)/sizeof(*input_b)/4
		== sizeof(input_c)/sizeof(*input_c)/5, "");

```
it also works in gcc:
```
test.c:5:1: error: static assertion failed: ""
    5 | _Static_assert(sizeof(input_a)/sizeof(*input_a)/3
      | ^~~~~~~~~~~~~~
```

SamuelMiksits:
> I tried:
> 
> ```c
> #include <stddef.h>
> 
> int input_a[10],
>     input_b[11],
>     input_c[12];
> 
> const size_t bufferSizeA = 3, bufferSizeB = 4, bufferSizeC = 5;
> 
> static const int iterations_a = sizeof(input_a)/sizeof(*input_a)/bufferSizeA;
> static const int iterations_b = sizeof(input_b)/sizeof(*input_b)/bufferSizeB;
> static const int iterations_c = sizeof(input_c)/sizeof(*input_c)/bufferSizeC;
> 
> _Static_assert(iterations_a == iterations_b == iterations_c, "");
> ```
> 
> This works in clang:
> 
> ```
> test.c:13:16: error: static assertion failed due to requirement 'iterations_a == iterations_b ==
>       iterations_c':
>    13 | _Static_assert(iterations_a == iterations_b == iterations_c, "");
>       |                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> test.c:13:45: note: expression evaluates to '0 == 2'
>    13 | _Static_assert(iterations_a == iterations_b == iterations_c, "");
>       |                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~
> 1 error generated.
> ```
> 
> but not in gcc:
> 
> ```
> test.c:13:45: error: expression in static assertion is not constant
>    13 | _Static_assert(iterations_a == iterations_b == iterations_c, "");
>       |                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~
> ```

Hmm okay. So I didn't realize it at first but yes of course `static_assert` (the normal one) is C++ not C. But you are using `_Static_assert`. I don't know why GCC and Clang would do that differently but does `input_a`, etc. all have to be `static const` for this to work on GCC?

klarasm:
https://en.cppreference.com/w/c/language/_Static_assert.html

Seems like `static_assert` can be used with C23, so `-std=c23` needed before gcc 15. Before C23, we are also supposed to `#include <assert.h>`

> Hmm okay. So I didn't realize it at first but yes of course `static_assert` (the normal one) is C++ not C. But you are using `_Static_assert`. I don't know why GCC and Clang would do that differently but does `input_a`, etc. all have to be `static const` for this to work on GCC?

That doesn't seem to help, unfortunately.

SamuelMiksits:
> https://en.cppreference.com/w/c/language/_Static_assert.html
> 
> Seems like `static_assert` can be used with C23, so `-std=c23` needed before gcc 15. Before C23, we are also supposed to `#include <assert.h>`

Okay does (your) GCC and Clang have different C standards by default? Also I guess the `bufferSizeA` etc was `const size_t` and not `static const size_t`, does that make a difference? Because if the second thing worked but not the first one that is the only that would be different?

klarasm:
> > https://en.cppreference.com/w/c/language/_Static_assert.html
> > Seems like `static_assert` can be used with C23, so `-std=c23` needed before gcc 15. Before C23, we are also supposed to `#include <assert.h>`
> 
> Okay does (your) GCC and Clang have different C standards by default? Also I guess the `bufferSizeA` etc was `const size_t` and not `static const size_t`, does that make a difference? Because if the second thing worked but not the first one that is the only that would be different?

I think actually clang still defaults to C17, but not sure. Adding `-std=c23` does not change gcc's behaviour.

using `static const` for bufferSize doesn't help either.

SamuelMiksits:
> I think actually clang still defaults to C17, but not sure. Adding `-std=c23` does not change gcc's behaviour.

Interesting, if I check my `man clang` it says:

```
The default C language standard is gnu17, except on PS4, where it is gnu99.
```

so `-std=gnu17` and not `-std=c17` but I can't find the default for gcc 
([docs](https://gcc.gnu.org/onlinedocs/gcc-14.3.0/gcc/Standards.html#C-Language) for GCC 14 which I have says
> The default, if no C language dialect options are given, is -std=gnu17. 

)

klarasm:
> > I think actually clang still defaults to C17, but not sure. Adding `-std=c23` does not change gcc's behaviour.
> 
> Interesting, if I check my `man clang` it says:
> 
> ```
> The default C language standard is gnu17, except on PS4, where it is gnu99.
> ```
> 
> so `-std=gnu17` and not `-std=c17` but I can't find the default for gcc

I believe GCC changed from gnu17 to gnu23 with GCC 15. Don't know when gnu17 became default.

So if I do:
```c
#include <stddef.h>

static const int input_a[10], input_b[11], input_c[12];

constexpr size_t bufferSizeA = 3;
constexpr size_t bufferSizeB = 4;
constexpr size_t bufferSizeC = 5;

constexpr int iterations_a = sizeof(input_a)/sizeof(*input_a)/bufferSizeA;
constexpr int iterations_b = sizeof(input_b)/sizeof(*input_b)/bufferSizeB;
constexpr int iterations_c = sizeof(input_c)/sizeof(*input_c)/bufferSizeC;

static_assert(iterations_a == iterations_b == iterations_c, "");
```
I can compile with `gcc -std=c23` and it works.

klarasm:
Though we can probably skip this for now. Even if it would be nice if we get to know if our tests are bad, it shouldn't block this.

SamuelMiksits:
> Though we can probably skip this for now. Even if it would be nice if we get to know if our tests are bad, it shouldn't block this.

I can maybe create an issue (should probably create an issue for this PR too lol) where it mentions #293 so we can go back and check what was explored here

SamuelMiksits:
I *probably* will need to go back to the documentation commit real quick and update that since I did move around a couple of things in the C code

klarasm (approved):
This looks good to me now!

SamuelMiksits:
Hmm thinking about it `io` should probably be in the `translationContext` instead of getting passed around in functions

src/ArgumentsMain.hs
```diff
@@ -36,14 +36,20 @@ data Target
   = PC
   | PICO2
 
+data IOType
```
sthaeron: `IOType` should probably be renamed to `InputType` as the current types only change how inputs are affected in the C code. If we were to start supporting different output types, we'd probably want to parse them separately and have a separate `OutputType` as well.

src/ArgumentsMain.hs
```diff
@@ -36,14 +36,20 @@ data Target
...
+  = Scanf
```
sthaeron: Perhaps the `scanf` option should just be called `stdin`. For the user, how a feature is implemented isn't important; it's the functionality that matters. `scanf` is very C-specific, so using an umbrella term might be more understandable.

src/ForSyDeIRToProceduralIR.hs
```diff
@@ -59,8 +62,17 @@ translateIRSystemToProgram dflags scheduleList bufferList delayBufferList lookup
 translateContextToMain :: TranslationContext -> [IRId] -> Global
 translateContextToMain context scheduleList =
   let scheduledStmts = scheduleActors context scheduleList
-      whileStmt = SWhile (EInt 1) (SScope scheduledStmts)
-      mainInitStmts = [SExpr (ECall "init" []), SVarDecl TInt "status"] ++ reverse (initBuffers context) ++ reverse (ioTokens context) ++ reverse (initDelay context)
+      scheduledStmtsSpecial = case (ioType context) of
```
sthaeron: Could you rename the variable `scheduledStmtsSpecial`? All variables in this module which use Procedural IR statements end in the suffix `Stmt` or `Stmts`; this one should too. Perhaps have `special` as a prefix or potentially rename it to `inputScheduledStmts`.

src/ForSyDeIRToProceduralIR.hs
```diff
@@ -59,8 +62,17 @@ translateIRSystemToProgram dflags scheduleList bufferList delayBufferList lookup
...
+        Scanf -> []
+        Predefined ->
+          [ SVarAssign "iter_current" (EBinOp Plus (EVar "iter_current") (EInt 1)),
+            SIf (EBinOp Equal (EVar "iter_current") (EVar "iter_max")) (SVarAssign "iter_current" (EInt 0)) Nothing
+          ]
+      whileStmt = SWhile (EInt 1) (SScope (scheduledStmts ++ scheduledStmtsSpecial))
+      mainInitStmtsSpecial = case (ioType context) of
```
sthaeron: Same here, perhaps `inputInitStmts`.

sthaeron:
I don’t think we should have the default state of using predefined data result in the compiled C program continuing indefinitely. Since the input sequence is predefined, we should just stop once the max number of iterations has been reached. Or we should have the user explicitly define a max number of runs in the `input.h` file.

docs/c-code-templates.md
```diff
@@ -222,6 +266,19 @@ all of the tokens have been printed, print a new line to end the schedule:
...
+
+If in `predefined` IO mode, it needs to be checked whether all the data has
+been consumed. Start by incrementing the iteration counter, and then wrap
+the value back to 0 if all inputs have been consumed:
+
+```C
+ iter_current = iter_current + 1;
+ if (iter_current == iter_max) {
+     iter_current = 0;
+ }
```
sthaeron, line 286: I suggest changing the generated C code to use `iteration_max` and `iteration_current`. There is no need to shorten the variable names.

sthaeron (changes requested):
All tests passed, I also tried out the new compiler feature and it seems to work as you've described. I've provided some comments for some minor tweaks. However, I do think we should change the default behaviour of this new feature to not run indefinitely.

src/ForSyDeIRToProceduralIR.hs
SamuelMiksits: I think "special" might be bad because if we need to add other paths of branching in the future then you would do have to do something stupid like "special2". I want to say that I don't like the idea of adding something as a prefix, but it might be because in my head I probably would want everything to have been `stmtsScheduled`, `stmtsMainInit`, in general. I just prefer it for searching/autocompleting to have the same thing at the start and then I can get a list of the variables and autocomplete it that way so we probably aren't going to agree on anything that both are happy with.

I think `input` is a bad prefix because we have some other code that depends on `ìnput`/`output` [signals] can we do something like `scheduledInputTypeStmts` and `mainInitInputTypeStmts`? Because where the code fits in it really belongs to either `scheduledStmts` but is then case'd depending on `inputType` or `mainInitStmts` that are case'd depending on `inputType`.

src/ArgumentsMain.hs
SamuelMiksits: Okay I agree with being able to split it in the future

src/ArgumentsMain.hs
SamuelMiksits: Hmm okay, I wanted explicitly to avoid using `StdIn` or something similar as a flag because we could end up in a scenario where we have 5 different argument types that all have something called `StdIn` which would be very confusing from a development perspective. But I can change it to `StdIn`

SamuelMiksits:
> I don’t think we should have the default state of using predefined data result in the compiled C program continuing indefinitely. Since the input sequence is predefined, we should just stop once the max number of iterations has been reached. Or we should have the user explicitly define a max number of runs in the `input.h` file.

Okay yeah it makes sense. If you feed some fixed length data into the SDF model in ForSyDe Shallow in Haskell it just runs until it can't produce any more outputs, and doesn't run perpetually. So if the default C path mimicked this behavior it would be reasonable. So then we probably need to add another flag `runs` and have `predefined` have this as the default behavior:
```C
if (iter_current == iter_max) {
    break;
}
```
and then with `runs` have either `runs={x}` (int value) mode with:

```C
if (iter_current == iter_max) {
    iter_current = 0;
    run_current += 1;
}
if (run_current == run_max) {
    break;
}
```

or `runs="inf"`

```C
if (iter_current == iter_max) {
    iter_current = 0;
}
```

src/ForSyDeIRToProceduralIR.hs
SamuelMiksits: Have changed variable names around and don't see shadow warnings anymore so I will mark as resolved

SamuelMiksits:
@sthaeron 
Now I have updated it so we have
- flag `--inputtype` with either `--inputtype=stdin` which returns `StdIn` or `--inputtype=predefined` which returns `Predefined`, `Predefined` by default
- flag `--runs`, with either `--runs=inf` that returns `Perpetual` or `--runs=x` which returns `(Limited x)`, where `(Limited 1)` is default.

This means you can run the program with the following configurations:
1. `--inputtype=stdin`, (possible to set `--runs` but pattern matching ignores these when `StdIn` is set
2. `--inputtype=predefined` `--runs=inf`
3. `--inputtype=predefined` `--runs=x`

I have done some testing on `SDF_example_008.hs` and you should get the following:

1. `--inputtype=stdin`
```C
#define PLATFORM PC
typedef int token;
#include "include/common.h"
#include <stdio.h>

static void add(int *input_1, int *input_2, int *output);

static void accumulate(int *input_1, int *input_2, int *output_1, int *output_2);

int main() {
    init();
    int status;
    buffer_nonblocking *s_in_y = buffer_nonblocking_new(1);
    buffer_nonblocking *s_in_x = buffer_nonblocking_new(1);
    buffer_nonblocking *s_out = buffer_nonblocking_new(1);
    buffer_nonblocking *s_1 = buffer_nonblocking_new(1);
    buffer_nonblocking *s_2_s_2_delayed = buffer_nonblocking_new(1);
    int input_s_in_y[1];
    int input_s_in_x[1];
    int output_s_out;
    write_token(s_2_s_2_delayed, 0);
    while (1) {
        for (int i = 0; i < 1; i++) {
            status = scanf("%d", &input_s_in_x[i]);
        }
        if (status < 1) break;
        for (int i = 0; i < 1; i++) {
            write_token(s_in_x, input_s_in_x[i]);
        }
        for (int i = 0; i < 1; i++) {
            status = scanf("%d", &input_s_in_y[i]);
        }
        if (status < 1) break;
        for (int i = 0; i < 1; i++) {
            write_token(s_in_y, input_s_in_y[i]);
        }
        actor21SDF(1, 1, 1, s_in_x, s_in_y, s_1, add);
        actor22SDF(1, 1, 1, 1, s_1, s_2_s_2_delayed, s_out, s_2_s_2_delayed, accumulate);
        for (int i = 0; i < 1; i++) {
            read_token(s_out, &output_s_out);
            printf("%d ", output_s_out);
        }
        printf("\n");
    }
    buffer_nonblocking_free(s_in_y);
    buffer_nonblocking_free(s_in_x);
    buffer_nonblocking_free(s_out);
    buffer_nonblocking_free(s_1);
    buffer_nonblocking_free(s_2_s_2_delayed);
    return 0;
}

static void add(int *input_1, int *input_2, int *output) {
    output[0] = input_1[0] + input_2[0];
}

static void accumulate(int *input_1, int *input_2, int *output_1, int *output_2) {
    output_1[0] = input_1[0] + input_2[0];
    output_2[0] = input_1[0] + input_2[0];
}

```

2. `--inputtype=predefined` `--runs=inf`

```C
#define PLATFORM PC
typedef int token;
#include "include/common.h"
#include <stdio.h>
#include "input.h"

static void add(int *input_1, int *input_2, int *output);

static void accumulate(int *input_1, int *input_2, int *output_1, int *output_2);

int main() {
    init();
    int iteration_current = 0;
    buffer_nonblocking *s_in_y = buffer_nonblocking_new(1);
    buffer_nonblocking *s_in_x = buffer_nonblocking_new(1);
    buffer_nonblocking *s_out = buffer_nonblocking_new(1);
    buffer_nonblocking *s_1 = buffer_nonblocking_new(1);
    buffer_nonblocking *s_2_s_2_delayed = buffer_nonblocking_new(1);
    int output_s_out;
    write_token(s_2_s_2_delayed, 0);
    while (1) {
        for (int i = 0; i < 1; i++) {
            write_token(s_in_x, input_s_in_x[iteration_current * 1 + i]);
        }
        for (int i = 0; i < 1; i++) {
            write_token(s_in_y, input_s_in_y[iteration_current * 1 + i]);
        }
        actor21SDF(1, 1, 1, s_in_x, s_in_y, s_1, add);
        actor22SDF(1, 1, 1, 1, s_1, s_2_s_2_delayed, s_out, s_2_s_2_delayed, accumulate);
        for (int i = 0; i < 1; i++) {
            read_token(s_out, &output_s_out);
            printf("%d ", output_s_out);
        }
        printf("\n");
        iteration_current = iteration_current + 1;
        if (iteration_current == iteration_max) {
            iteration_current = 0;
        }
    }
    buffer_nonblocking_free(s_in_y);
    buffer_nonblocking_free(s_in_x);
    buffer_nonblocking_free(s_out);
    buffer_nonblocking_free(s_1);
    buffer_nonblocking_free(s_2_s_2_delayed);
    return 0;
}

static void add(int *input_1, int *input_2, int *output) {
    output[0] = input_1[0] + input_2[0];
}

static void accumulate(int *input_1, int *input_2, int *output_1, int *output_2) {
    output_1[0] = input_1[0] + input_2[0];
    output_2[0] = input_1[0] + input_2[0];
}

```

3. `--inputtype=predefined` (defaults to `--runs=1`)

```C
#define PLATFORM PC
typedef int token;
#include "include/common.h"
#include <stdio.h>
#include "input.h"

static void add(int *input_1, int *input_2, int *output);

static void accumulate(int *input_1, int *input_2, int *output_1, int *output_2);

int main() {
    init();
    int iteration_current = 0;
    int run_current = 0;
    int run_max = 1;
    buffer_nonblocking *s_in_y = buffer_nonblocking_new(1);
    buffer_nonblocking *s_in_x = buffer_nonblocking_new(1);
    buffer_nonblocking *s_out = buffer_nonblocking_new(1);
    buffer_nonblocking *s_1 = buffer_nonblocking_new(1);
    buffer_nonblocking *s_2_s_2_delayed = buffer_nonblocking_new(1);
    int output_s_out;
    write_token(s_2_s_2_delayed, 0);
    while (1) {
        for (int i = 0; i < 1; i++) {
            write_token(s_in_x, input_s_in_x[iteration_current * 1 + i]);
        }
        for (int i = 0; i < 1; i++) {
            write_token(s_in_y, input_s_in_y[iteration_current * 1 + i]);
        }
        actor21SDF(1, 1, 1, s_in_x, s_in_y, s_1, add);
        actor22SDF(1, 1, 1, 1, s_1, s_2_s_2_delayed, s_out, s_2_s_2_delayed, accumulate);
        for (int i = 0; i < 1; i++) {
            read_token(s_out, &output_s_out);
            printf("%d ", output_s_out);
        }
        printf("\n");
        iteration_current = iteration_current + 1;
        if (iteration_current == iteration_max) {
            iteration_current = 0;
            run_current = run_current + 1;
        }
        if (run_current == run_max) {
            break;
        }
    }
    buffer_nonblocking_free(s_in_y);
    buffer_nonblocking_free(s_in_x);
    buffer_nonblocking_free(s_out);
    buffer_nonblocking_free(s_1);
    buffer_nonblocking_free(s_2_s_2_delayed);
    return 0;
}

static void add(int *input_1, int *input_2, int *output) {
    output[0] = input_1[0] + input_2[0];
}

static void accumulate(int *input_1, int *input_2, int *output_1, int *output_2) {
    output_1[0] = input_1[0] + input_2[0];
    output_2[0] = input_1[0] + input_2[0];
}

```

SamuelMiksits:
Oh one sec I have to rewrite the documentation for the 7th time

src/ForSyDeIRToProceduralIR.hs
sthaeron: I agree that our naming convention is backwards, but for now, it's best to just keep things consistent. The naming changes you made to variables seemed good.

src/ForSyDeIRToProceduralIR.hs
```diff
@@ -153,25 +189,37 @@ translateIRConstructor initialContext constructor = case constructor of
...
-              breakIfStmt = SIf (EBinOp Less (EVar "status") (EInt 1)) (SBreak) Nothing
-              writeForStmt =
-                SFor
-                  (SVarDef TInt "i" (EInt 0))
-                  (EBinOp Less (EVar "i") (EInt (bufferSize)))
-                  (SExpr (EUnOp Increment (EVar "i")))
-                  (SScope [SExpr (ECall "write_token" [EVar $ show signalId, EArrayAccess (EVar ("input_" ++ show signalId)) (EVar "i")])])
-           in (writeForStmt : breakIfStmt : scanForStmt : acc)
+        then case (inputType initialContext) of
+          -- If IO-Type is scanf, then do scanf->break->write
```
sthaeron: Outdated comment should be `InputType`.

src/ForSyDeIRToProceduralIR.hs
```diff
@@ -153,25 +189,37 @@ translateIRConstructor initialContext constructor = case constructor of
...
+                    (SScope [SVarAssign "status" (ECall "scanf" [EString "%d", EReference (EArrayAccess (EVar ("input_" ++ show signalId)) (EVar "i"))])])
+                breakIfStmt = SIf (EBinOp Less (EVar "status") (EInt 1)) (SBreak) Nothing
+                writeForStmt =
+                  SFor
+                    (SVarDef TInt "i" (EInt 0))
+                    (EBinOp Less (EVar "i") (EInt (bufferSize)))
+                    (SExpr (EUnOp Increment (EVar "i")))
+                    (SScope [SExpr (ECall "write_token" [EVar $ show signalId, EArrayAccess (EVar ("input_" ++ show signalId)) (EVar "i")])])
+             in (writeForStmt : breakIfStmt : scanForStmt : acc)
+          -- If IO-Type is predefined, then do write
```
sthaeron: Outdated comment should be `InputType`.

src/ForSyDeIRToProceduralIR.hs
```diff
@@ -59,8 +64,33 @@ translateIRSystemToProgram dflags scheduleList bufferList delayBufferList lookup
...
+          --  SIf (EBinOp Equal (EVar "iteration_current") (EVar "iteration_max")) (SVarAssign "iteration_current" (EInt 0)) Nothing,
+          --  SVarAssign "run_current" (EBinOp Plus (EVar "run_current") (EInt 1))
+          -- ]
+          [ SVarAssign "iteration_current" (EBinOp Plus (EVar "iteration_current") (EInt 1)),
+            SIf (EBinOp Equal (EVar "iteration_current") (EVar "iteration_max")) (SScope [SVarAssign "iteration_current" (EInt 0), SVarAssign "run_current" (EBinOp Plus (EVar "run_current") (EInt 1))]) Nothing
+          ]
+      scheduledRunsStmts = case ((inputType context), (runs context)) of
+        (StdIn, _) -> []
+        (Predefined, Perpetual) -> []
+        (Predefined, (Limited _)) -> [SIf (EBinOp Equal (EVar "run_current") (EVar "run_max")) (SScope [SBreak]) Nothing]
```
sthaeron, line 84: Could probably have a base case the default `runs=1` case `(Predefined, (Limited 1)) -> []`

src/ForSyDeIRToProceduralIR.hs
```diff
@@ -59,8 +64,33 @@ translateIRSystemToProgram dflags scheduleList bufferList delayBufferList lookup
...
   let scheduledStmts = scheduleActors context scheduleList
-      whileStmt = SWhile (EInt 1) (SScope scheduledStmts)
-      mainInitStmts = [SExpr (ECall "init" []), SVarDecl TInt "status"] ++ reverse (initBuffers context) ++ reverse (ioTokens context) ++ reverse (initDelay context)
+      scheduledInputStmts = case ((inputType context), (runs context)) of
+        (StdIn, _) -> []
+        (Predefined, Perpetual) ->
+          [ SVarAssign "iteration_current" (EBinOp Plus (EVar "iteration_current") (EInt 1)),
+            SIf (EBinOp Equal (EVar "iteration_current") (EVar "iteration_max")) (SScope [SVarAssign "iteration_current" (EInt 0)]) Nothing
+          ]
+        (Predefined, Limited _) ->
```
sthaeron, line 73: Could probably have a base case for the default `runs=1` case 
```haskell
(Predefined, (Limited 1)) -> [
    SVarAssign "iteration_current" (EBinOp Plus (EVar "iteration_current") (EInt 1)),
    SIf (EBinOp Equal (EVar "iteration_current") (EVar "iteration_max")) (SScope [SBreak]) Nothing
]
```

sthaeron (approved):
I've tested the new and updated features you've added and it all seems to be working great. I've provided a few comments on the code, but nothing that should prevent you from merging. I've only got a little nitpick, could you rename the argument flag `--inputtype` to `--input-type`, since you've always used dashes between words for all the other argument options.

src/ForSyDeIRToProceduralIR.hs
SamuelMiksits: I see your point. I tried compiling with `-O2` when I have `int run_max = 1;` and then I get the following main:

```
  40135a:       e8 f1 fc ff ff          call   401050 <printf@plt>
  40135f:       bf 0a 00 00 00          mov    $0xa,%edi
  401364:       e8 d7 fc ff ff          call   401040 <putchar@plt>
  401369:       83 c3 01                add    $0x1,%ebx
  40136c:       39 1d be 2c 00 00       cmp    %ebx,0x2cbe(%rip)        # 404030 <iteration_max>
  401372:       0f 85 15 ff ff ff       jne    40128d <main+0x64>
  401378:       4c 89 ff                mov    %r15,%rdi
  40137b:       e8 92 fe ff ff          call   401212 <buffer_nonblocking_free>
```
and then for `int run_max = 10000;`
```
  401364:       e8 e7 fc ff ff          call   401050 <printf@plt>
  401369:       bf 0a 00 00 00          mov    $0xa,%edi
  40136e:       e8 cd fc ff ff          call   401040 <putchar@plt>
  401373:       83 c3 01                add    $0x1,%ebx
  401376:       39 1d b4 2c 00 00       cmp    %ebx,0x2cb4(%rip)        # 404030 <iteration_max>
  40137c:       0f 85 14 ff ff ff       jne    401296 <main+0x6d>
  401382:       83 44 24 1c 01          addl   $0x1,0x1c(%rsp)
  401387:       8b 44 24 1c             mov    0x1c(%rsp),%eax
  40138b:       3d 10 27 00 00          cmp    $0x2710,%eax
  401390:       74 0a                   je     40139c <main+0x173>
  401392:       bb 00 00 00 00          mov    $0x0,%ebx
  401397:       e9 fa fe ff ff          jmp    401296 <main+0x6d>
  40139c:       4c 89 ff                mov    %r15,%rdi
  40139f:       e8 6e fe ff ff          call   401212 <buffer_nonblocking_free>
```
so it clearly looks like the compiler is already doing this optimization for us and I don't think we should worry about this.

klarasm:
@SamuelMiksits,
Btw, just saw this now.
```c
    while (1) {
        iter_current = iter_current + 1;
        if (iter_current == iter_max) {
            iter_current = 0;
        }
        for (int i = 0; i < 1; i++) {
            write_token(s_in_x, input_s_in_x[iter_current * 1 + i]);
        }
        for (int i = 0; i < 1; i++) {
            write_token(s_in_y, input_s_in_y[iter_current * 1 + i]);
        }
        actor21SDF(1, 1, 1, s_in_x, s_in_y, s_1, add);
        actor22SDF(1, 1, 1, 1, s_1, s_2_s_2_delayed, s_out, s_2_s_2_delayed, accumulate);
        for (int i = 0; i < 1; i++) {
            read_token(s_out, &output_s_out);
            printf("%d", output_s_out);
        }
        printf("\n");
    }
```
You're using `i` to index the input arrays which is wrong. It's zeroed in each loop. We need a separate counter  for each input.

klarasm:
Also, this now currently multiplies `iter_current` by the length of the loop. This does not work when there are multiple loops for the same input.

SamuelMiksits:
> @SamuelMiksits, Btw, just saw this now.
> 
> ```c
>     while (1) {
>         iter_current = iter_current + 1;
>         if (iter_current == iter_max) {
>             iter_current = 0;
>         }
>         for (int i = 0; i < 1; i++) {
>             write_token(s_in_x, input_s_in_x[iter_current * 1 + i]);
>         }
>         for (int i = 0; i < 1; i++) {
>             write_token(s_in_y, input_s_in_y[iter_current * 1 + i]);
>         }
>         actor21SDF(1, 1, 1, s_in_x, s_in_y, s_1, add);
>         actor22SDF(1, 1, 1, 1, s_1, s_2_s_2_delayed, s_out, s_2_s_2_delayed, accumulate);
>         for (int i = 0; i < 1; i++) {
>             read_token(s_out, &output_s_out);
>             printf("%d", output_s_out);
>         }
>         printf("\n");
>     }
> ```
> 
> You're using `i` to index the input arrays which is wrong. It's zeroed in each loop. We need a separate counter for each input.

> Also, this now currently multiplies `iter_current` by the length of the loop. This does not work when there are multiple loops for the same input.


I implemented this on SDF_example_002 by hand initially:

```C
// same input as simulation.md
token input_a[7*4] = {1, 1, 2, 4, 2, 2, 2, 2, 4, 1, 2, 1, 1, 1, 1, 1,
        0, 0, 0, 0, 0, 0, 0, 0, 1, 5, 2, 3};
token input_b[7] = {1, 2, 1, 1, 0, 0, 4};

int iter_max = 7;

while(1) {
    iter_current += 1;
    if (iter_current == iter_max) {
        iter_current = 0;
    }

    // Write inputs to buffer(s)
    for(i = 0; i < 4; i++) {
        write_token(s_in_a, input_a[iter_current*4 + i]);
    }

    for(i = 0; i < 1; i++) {
        write_token(s_in_b, input_b[iter_current*1 + i]);
    }
```
and produced the same output as the simulation:
```
Output: 3 11 
Output: 5 24 
Output: 3 35 
Output: 3 42 
Output: 1 43 
Output: 1 44 
Output: 9 64 
```
I don't really understand what the problem with the array flattening is

klarasm:
Here is the output when run on `SDF_example_007.hs` (compiled with Sebastian's `coreToProcedural` branch):
```c
    while (1) {
        for (int i = 0; i < 2; i++) {
            write_token(s_in_a, input_s_in_a[iteration_current * 2 + i]);
        }
        actor12SDF(2, 2, 1, s_in_a, s_1, s_2, f_1);
        for (int i = 0; i < 2; i++) {
            write_token(s_in_a, input_s_in_a[iteration_current * 2 + i]);
        }
        actor12SDF(2, 2, 1, s_in_a, s_1, s_2, f_1);
        for (int i = 0; i < 1; i++) {
            write_token(s_in_b, input_s_in_b[iteration_current * 1 + i]);
        }
        actor13SDF(1, 2, 1, 2, s_in_b, s_3, s_4, s_5, f_2);
        for (int i = 0; i < 1; i++) {
            write_token(s_in_b, input_s_in_b[iteration_current * 1 + i]);
        }
        actor13SDF(1, 2, 1, 2, s_in_b, s_3, s_4, s_5, f_2);
        actor32SDF(2, 4, 2, 2, 2, s_2, s_3, s_4, s_6, s_7, f_3);
        actor41SDF(4, 2, 2, 4, 2, s_1, s_6, s_7, s_5, s_out, f_4);
        for (int i = 0; i < 2; i++) {
            read_token(s_out, &output_s_out);
            printf("%d ", output_s_out);
        }
        printf("\n");
        iteration_current = iteration_current + 1;
        if (iteration_current == iteration_max) {
            iteration_current = 0;
            run_current = run_current + 1;
        }
        if (run_current == run_max) {
            break;
        }
    }
```
We have two loops for each input, but they will read the same elements as the first loop. Also note that the total input consumption rate of `input_s_in_a` is 4 rather than 2 and `input_s_in_b` is 2 rather than 1. This means that we also will get the wrong offset in the next iteration of the schedule.

SamuelMiksits:
@klarasm Okay wait. So we decided (and as such I have written in the documentation) that:

> - The input tokens should be of type `token *`, and a unique array needs to be created
for every input in the graph. The inputs will need to be processed for the whole period
at once, which means that the (static) length of each array will need to be equal to
tokens consumed per firing, multiplied by that actors value in the repetition vector.
The input arrays should **only** be declared in `stdin` IO mode, and should be ommitted
in the `predefined` mode since they are already defined in the `input.h` file.
>
>Example:
>
>- `input_a` input for actor A. A consumes 2 tokens from input each time, and is ran 2
times per schedule: `2*2 -> 4`
>- `input_b` input for actor B. B consumes 1 token from input each time, and is ran 1
time per schedule.
>
> ```C
> // Temporary tokens for print/scan functions for input and output
> token input_a[4];
> token input_b[1];
> token output;
> ```
but looking at the actual code we have

```hs
let bufferSize = getTargetRate initialContext signalId
                -- If actor has inputs which are system inputs the following
                -- adds statements which relate to obtaining inputs from
                -- standard in.
                scanForStmt =
                  SFor
                    (SVarDef TInt "i" (EInt 0))
                    (EBinOp Less (EVar "i") (EInt (bufferSize)))
                    (SExpr (EUnOp Increment (EVar "i")))
                    (SScope [SVarAssign "status" (ECall "scanf" [EString "%d", EReference (EArrayAccess (EVar ("input_" ++ show signalId)) (EVar "i"))])])
```
and `getTargetRate` only returns the rate from the IR signal and does not multiply by how many times the actor is ran as. So I don't understand how it is handled when you need to read multiple times for an actor, is the expectation that the for-loop needs to be ran multiple times instead of allocating the buffer size needed for the whole period and reads all input at once? I have only been able to test the compiler on 008 so have not had the opportunity to check this.

klarasm:
Yes, basically we can't do this now that we don't read the input all at once. You should be able to check now on Sebastians coreToProcedural which will produce output for all our examples.

SamuelMiksits:
Okay, I had literally no idea we decided to not read all inputs at once

klarasm:
It was in https://github.com/sthaeron/forsyde-devtools/pull/264#discussion_r2556954062 and https://github.com/sthaeron/forsyde-devtools/pull/264#issuecomment-3581038785

I thought we might have talked about it on a standup/meeting but I'm probably misremembering. Though neither @sthaeron or I thought to update the documentation, sorry about that.

SamuelMiksits:
> It was in [#264 (comment)](https://github.com/sthaeron/forsyde-devtools/pull/264#discussion_r2556954062) and [#264 (comment)](https://github.com/sthaeron/forsyde-devtools/pull/264#issuecomment-3581038785)
> 
> I thought we might have talked about it on a standup/meeting but I'm probably misremembering. Though neither @sthaeron or I thought to update the documentation, sorry about that.

No worries, I think I must have missed something in review or communication. I guess I will have to figure something out

SamuelMiksits:
I have one idea, how about we declare `i_a, i_b, i_...`, for every actor at the start. Then we reset those variables at the end of the loop as such:

```C
    token input_a[7*4] = {1, 1, 2, 4, 2, 2, 2, 2, 4, 1, 2, 1, 1, 1, 1, 1,
        0, 0, 0, 0, 0, 0, 0, 0, 1, 5, 2, 3};
    token input_b[7] = {1, 2, 1, 1, 0, 0, 4};

	//int runs = 1E6;
    int runs = 1;
	int run_current = 0;
    int iteration_current = 0;
    int iteration_max = 7;
    int i_a = 0;
    int i_b = 0;

    // Repeating Schedule: A A B C D
    while(1) {
        int ret;

	    // Write inputs to buffer(s)
	    for(i = 0; i < 2; i++) {
            write_token(s_in_a, input_a[iteration_current*4 + i_a]);
            i_a = i_a + 1;
        }

	    // A
	    actor11SDF(2, 1, s_in_a, a_1, f_a);

        // Write inputs to buffer(s)
	    for(i = 0; i < 2; i++) {
            write_token(s_in_a, input_a[iteration_current*4 + i_a]);
            i_a = i_a + 1;
        }
	    // A
	    actor11SDF(2, 1, s_in_a, a_1, f_a);

	    for(i = 0; i < 1; i++) {
            write_token(s_in_b, input_b[iteration_current*1 + i_b]);
            i_b = i_b + 1;
        }
	    // B
	    actor11SDF(1, 2, s_in_b, a_2, f_b);

	    // C
	    actor21SDF(2, 1, 1, a_1, a_4, a_3, f_c);

	    // D
	    actor22SDF(2, 1, 1, 2, a_2, a_3, a_4, s_out, f_d);

	    // Write output tokens
	    printf("Output: ");
	    for(i = 0; i < 2; i++) {
	        read_token(s_out, &output);
	        printf("%d ", output);
	    }
	    printf("\n");
        i_a = 0;
        i_b = 0;

        iteration_current = iteration_current + 1;
		if (iteration_current == iteration_max) {
            iteration_current = 0;
            // Reset all array subindexing variables
            run_current = run_current + 1;
        }

		if (run_current == runs) {
            break;
        }

    }
```
this still has the problem of requiring 
```C
// Write inputs to buffer(s)
	    for(i = 0; i < 2; i++) {
            write_token(s_in_a, input_a[iteration_current*4 + i_a]);
            i_a = i_a + 1;
        }
```
this `4` in the loop which is from how many tokens you need per firing and how many times it is ran in the schedule, but I can't think of anything else that you can do that is non-invasive to our current code structure apart from going in and making it so we only read input once.

I suppose we have

```hs
-- | Returns a list of statements corresponding to actor function calls based
-- on a provided Schedule. Note that actor function call statements include
-- the associated minimum runtime enviorment statements for interacting with
-- standard in and out.
scheduleActors :: TranslationContext -> [IRId] -> [Statement]
scheduleActors context scheduleList = foldl' foldSchedule [] scheduleList
  where
    foldSchedule :: [Statement] -> IRId -> [Statement]
    foldSchedule acc actorId =
      let actorStmts = case lookup actorId (actors context) of
            Just stmts -> stmts
            Nothing -> error ("schedule - actor not found: " ++ show actorId)
       in (acc ++ actorStmts)
```
can we check `acc` for how many times each actor has fired in this function and then back-calculate this number? Can otherwise make it a tuple type that does both and then just throw away the other thing and return what we want

SamuelMiksits:
> Here is the output when run on `SDF_example_007.hs` (compiled with Sebastian's `coreToProcedural` branch):
> 
> ```c
>     while (1) {
>         for (int i = 0; i < 2; i++) {
>             write_token(s_in_a, input_s_in_a[iteration_current * 2 + i]);
>         }
>         actor12SDF(2, 2, 1, s_in_a, s_1, s_2, f_1);
>         for (int i = 0; i < 2; i++) {
>             write_token(s_in_a, input_s_in_a[iteration_current * 2 + i]);
>         }
>         actor12SDF(2, 2, 1, s_in_a, s_1, s_2, f_1);
>         for (int i = 0; i < 1; i++) {
>             write_token(s_in_b, input_s_in_b[iteration_current * 1 + i]);
>         }
>         actor13SDF(1, 2, 1, 2, s_in_b, s_3, s_4, s_5, f_2);
>         for (int i = 0; i < 1; i++) {
>             write_token(s_in_b, input_s_in_b[iteration_current * 1 + i]);
>         }
>         actor13SDF(1, 2, 1, 2, s_in_b, s_3, s_4, s_5, f_2);
>         actor32SDF(2, 4, 2, 2, 2, s_2, s_3, s_4, s_6, s_7, f_3);
>         actor41SDF(4, 2, 2, 4, 2, s_1, s_6, s_7, s_5, s_out, f_4);
>         for (int i = 0; i < 2; i++) {
>             read_token(s_out, &output_s_out);
>             printf("%d ", output_s_out);
>         }
>         printf("\n");
>         iteration_current = iteration_current + 1;
>         if (iteration_current == iteration_max) {
>             iteration_current = 0;
>             run_current = run_current + 1;
>         }
>         if (run_current == run_max) {
>             break;
>         }
>     }
> ```
> 
> We have two loops for each input, but they will read the same elements as the first loop. Also note that the total input consumption rate of `input_s_in_a` is 4 rather than 2 and `input_s_in_b` is 2 rather than 1. This means that we also will get the wrong offset in the next iteration of the schedule.

Also if you look at SDF_example_007 from this branch it has:

```C
  buffer_nonblocking *s_in_b = buffer_nonblocking_new(2);
  buffer_nonblocking *s_in_a = buffer_nonblocking_new(4);
```
so the scheduler has calculated for the whole period assuming all inputs are read at once.

klarasm:
I mean, if it's easier to change the generation so it does what we wanted in the beginning, I'm not opposed.

PR #303, opened 2025-12-09
Core IR to Procedural IR

sthaeron:
Adds functionality to translate IRFunctions from ForSyDe IR into Procedural IR.
- Adds a module which translates Core IR function expressions into Procedural IR
- Adds various SDF examples used for testing
- Makes C the default output of the compiler

### DIFF
--- a/examples/model/SDF_example_011.hs
+++ b/examples/model/SDF_example_011.hs
@@ -1,0 +1,17 @@
+module SDF_example_011 where
+
+import ForSyDe.Shallow
+
+-- Net list
+system :: Signal Int -> Signal Int
+system s_in = s_out
+  where
+    s_out = a_a s_in
+
+-- Process specifications
+a_a :: Signal Int -> Signal Int
+a_a s = actor11SDF 1 1 add s
+
+-- Function definitions
+add :: [Int] -> [Int]
+add [x] = [x + 1]
--- a/examples/model/SDF_example_011.hs
+++ b/examples/model/SDF_example_011.hs
@@ -2,7 +2,7 @@ module SDF_example_011 where
 
 import ForSyDe.Shallow
 
--- Net list
+-- Netlist
 system :: Signal Int -> Signal Int
 system s_in = s_out
   where
--- a/src/CoreIR.hs
+++ b/src/CoreIR.hs
@@ -2,16 +2,22 @@ module CoreIR where
 
 import Data.List (intercalate)
 import GHC
+import GHC.Builtin.Uniques (knownUniqueName)
 import GHC.Core
 import GHC.Driver.Ppr
-import GHC.Plugins (Literal, Var (varName), litValue, occName, occNameString)
+import GHC.Plugins (Literal, Var (varName), litValue, nameStableString, occName, occNameString, varUnique)
 import Text.Printf (printf)
 
 indent :: String -> String
 indent = unlines . map ("  " ++) . lines
 
 varToString :: Var -> String
-varToString var = occNameString $ occName $ varName var
+varToString var =
+  let stringName = occNameString $ occName $ varName var
+      maybeUniqueName = knownUniqueName $ varUnique var
+   in case maybeUniqueName of
+        Just uniqueName -> stringName ++ nameStableString uniqueName
+        Nothing -> stringName
 
 literalToInt :: Literal -> Int
 literalToInt literal = fromIntegral $ litValue literal
@@ -29,13 +35,13 @@ prettyCoreProgram dflags = intercalate "\n" . map (prettyCoreBind dflags)
 
 prettyCoreExpr :: DynFlags -> CoreExpr -> String
 prettyCoreExpr dflags expr = case expr of
-  Var i -> printf "Var(%s)" (varToString i)
+  Var i -> printf "Var(%s)" (showPpr dflags i)
   Lit l -> printf "Lit(%s)" (showPpr dflags l)
   App e a -> printf "App(%s *\n%s)" (prettyCoreExpr dflags e) (prettyCoreExpr dflags a)
-  Lam b e -> printf "Lam(%s ->\n%s)" (varToString b) (prettyCoreExpr dflags e)
+  Lam b e -> printf "Lam(%s ->\n%s)" (showPpr dflags b) (prettyCoreExpr dflags e)
   Type t -> printf "Type(%s)" (showPpr dflags t)
   Let bind e -> printf "Let(\n%s in\n%s)" (indent (prettyCoreBind dflags bind)) (indent (prettyCoreExpr dflags e))
-  Case e b _ alts -> printf "Case(%s of %s {\n%s})" (prettyCoreExpr dflags e) (showPpr dflags b) (indent (prettyCoreAltList dflags alts))
+  Case e _ t alts -> printf "Case(%s of %s {\n%s})" (prettyCoreExpr dflags e) (showPpr dflags t) (indent (prettyCoreAltList dflags alts))
   Cast e co -> printf "Cast(%s by %s)" (prettyCoreExpr dflags e) (showPpr dflags co)
   Tick t e -> printf "Tick(%s: %s)" (showPpr dflags t) (prettyCoreExpr dflags e)
   Coercion co -> printf "Coercion(%s)" (showPpr dflags co)
--- a/src/CoreIRToProceduralIR.hs
+++ b/src/CoreIRToProceduralIR.hs
@@ -1,0 +1,92 @@
+{-# LANGUAGE DataKinds #-}
+
+module CoreIRToProceduralIR where
+
+import CoreIR (literalToInt, prettyCoreExpr, varToString)
+import Data.Maybe (listToMaybe, mapMaybe)
+import ForSyDeIR
+import GHC hiding (targetId)
+import GHC.Core
+import ProceduralIR
+
+data TranslationContext = TranslationContext
+  { flags :: DynFlags,
+    inputSignalIds :: [String],
+    outputSignalIds :: [String]
+  }
+
+initialTranslationContext :: DynFlags -> TranslationContext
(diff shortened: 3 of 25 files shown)


SamuelMiksits:
```C
for (int i = 0; i < 2; i++) {
    status = scanf("%d", &input_s_in_1[i]);
}
if (status < 1) break;
for (int i = 0; i < 2; i++) {
    write_token(s_in_1, input_s_in_1[i]);
}
```
oopsie ("""someone""" changed if so that it now puts break on the same line)

SamuelMiksits:
> ```c
> for (int i = 0; i < 2; i++) {
>     status = scanf("%d", &input_s_in_1[i]);
> }
> if (status < 1) break;
> for (int i = 0; i < 2; i++) {
>     write_token(s_in_1, input_s_in_1[i]);
> }
> ```
> 
> oopsie ("""someone""" changed if so that it now puts break on the same line)

so basically this
```diff
diff --git a/src/ForSyDeIRToProceduralIR.hs b/src/ForSyDeIRToProceduralIR.hs
index e3b0399..df4c4d0 100644
--- a/src/ForSyDeIRToProceduralIR.hs
+++ b/src/ForSyDeIRToProceduralIR.hs
@@ -203,7 +203,7 @@ translateIRConstructor initialContext constructor = case constructor of
                     (EBinOp Less (EVar "i") (EInt (bufferSize)))
                     (SExpr (EUnOp Increment (EVar "i")))
                     (SScope [SVarAssign "status" (ECall "scanf" [EString "%d", EReference (EArrayAccess (EVar ("input_" ++ show signalId)) (EVar "i"))])])
-                breakIfStmt = SIf (EBinOp Less (EVar "status") (EInt 1)) (SBreak) Nothing
+                breakIfStmt = SIf (EBinOp Less (EVar "status") (EInt 1)) (SScope [SBreak]) Nothing
                 writeForStmt =
                   SFor
                     (SVarDef TInt "i" (EInt 0))
```

examples/model/SDF_example_011.hs
```diff
@@ -0,0 +1,17 @@
+module SDF_example_011 where
+
+import ForSyDe.Shallow
+
+-- Net list
```
SamuelMiksits: I would say we are about 50/50 having
```hs
-- Netlist
```
and
```hs
-- Net list
```
Should we do something about this?

src/ForSyDeIRToProceduralIR.hs
```diff
@@ -56,7 +57,7 @@ translateIRSystemToProgram dflags scheduleList bufferList delayBufferList lookup
   let initialContext = initialTranslationContext dflags input r lookupSignalList inputList outputList delayBufferList
       context1 = foldl' translateIRConstructor initialContext constructors
       context2 = foldl' translateBuffer context1 bufferList
-      context3 = foldl' translateIRFunctionToGlobals context2 functionList
+      context3 = foldl' (translateIRFunctionToGlobals constructors) context2 functionList
```
SamuelMiksits, line 60: Might be fine here since it has minimal impact, but I moved variables like `runs` and `inputType` into translation context because otherwise you would need to perform massive partial applications for the fold, i.e:

```hs
context3 = foldl' (translateIRFunctionToGlobals constructors runs inputType) context2 functionList
```

and this defeats the purpose of having translation context in the first place

src/CoreIRToProceduralIR.hs
```diff
@@ -0,0 +1,277 @@
...
+  -- Function with 2 inputs
+  Lam b1 (Lam b2 e) ->
+    let context1 = context {functionInputs = [(IRVar b1), (IRVar b2)]}
+     in translateFunctionExpr context1 inputIndex inputCounter e
+  -- Function with 1 input
+  Lam b e ->
+    let context1 = context {functionInputs = [(IRVar b)]}
+     in translateFunctionExpr context1 inputIndex inputCounter e
+  Let _ e -> translateFunctionExpr context inputIndex inputCounter e
+  Case (Var i) _ _ty alts -> case elemIndex (IRVar i) (functionInputs context) of
```
SamuelMiksits, line 141: It is clear from the other comments when the cases are invoked, but what is this case?

SamuelMiksits (commented):
I think it looks fine in general. Could just do a few more comments in some places.

src/CoreIRToProceduralIR.hs
```diff
@@ -0,0 +1,277 @@
...
+getFunctionDefinition :: TranslationContext -> [(ProceduralIR.Type, String)] -> IRId -> Global
+getFunctionDefinition context parametersList functionId =
+  let stmtScope = scope context
+      functionDefinitionGlobal = GFuncDef (Just Static) TVoid (show functionId) parametersList stmtScope
+   in functionDefinitionGlobal
+
+findActorFromFunctionId :: IRId -> [IRConstructor] -> Maybe IRConstructor
+findActorFromFunctionId targetFunctionId constructors =
+  let checkIRConstructor :: IRConstructor -> Maybe IRConstructor
+      checkIRConstructor actor@(IRActor _ _ functionId _)
```
SamuelMiksits: I think the usage of the "@" here and in FPlus, etc is pretty esoteric and non-obvious and warrants some explanation

src/CoreIR.hs
```diff
@@ -29,13 +35,13 @@ prettyCoreProgram dflags = intercalate "\n" . map (prettyCoreBind dflags)
 
 prettyCoreExpr :: DynFlags -> CoreExpr -> String
 prettyCoreExpr dflags expr = case expr of
-  Var i -> printf "Var(%s)" (varToString i)
+  Var i -> printf "Var(%s)" (showPpr dflags i)
   Lit l -> printf "Lit(%s)" (showPpr dflags l)
   App e a -> printf "App(%s *\n%s)" (prettyCoreExpr dflags e) (prettyCoreExpr dflags a)
-  Lam b e -> printf "Lam(%s ->\n%s)" (varToString b) (prettyCoreExpr dflags e)
+  Lam b e -> printf "Lam(%s ->\n%s)" (showPpr dflags b) (prettyCoreExpr dflags e)
```
klarasm, line 41: Is there a reason we go back to showPpr here? Is it he unique stuff (you add a varUnique above)?

src/CoreIRToProceduralIR.hs
```diff
@@ -0,0 +1,277 @@
...
+  | FDiv
+  | FNegate
+
+translateIRFunction :: IRFunction -> DynFlags -> [IRConstructor] -> (Global, Maybe Global)
+translateIRFunction function dflags constructors =
+  let context = initialTranslationContext dflags
+   in case function of
+        IRFunction functionId Nothing ->
+          let actor = case (findActorFromFunctionId functionId constructors) of
+                Nothing -> error ("translateIRFunction - no actor associated to " ++ show functionId)
```
klarasm, line 63: Is there a reason we error if there's no actor associated here and below? As we're making them static it would just not be in the binary. Would it be possible to just process it normally and don't crash the compiler on it?

src/CoreIRToProceduralIR.hs
```diff
@@ -0,0 +1,277 @@
...
+                Just a -> a
+              (parameterList, _context1) = getParameterList context actor
+              (functionDeclarationGlobal) = getFunctionDeclaration parameterList functionId
+           in (functionDeclarationGlobal, Nothing)
+        IRFunction functionId (Just functionExpr) ->
+          let actor = case (findActorFromFunctionId functionId constructors) of
+                Nothing -> error ("translateIRFunction - no actor associated to " ++ show functionId)
```
klarasm, line 70: Here too.

src/ForSyDeIR.hs
```diff
@@ -42,6 +43,7 @@ data IRId
   | Empty
 
 instance Show IRId where
+  show :: IRId -> String
```
klarasm, line 46: This seems a little unrelated, any reason for it?

klarasm (changes requested):
Looks good in general, but do we really need to crash if we can't associate a function to an actor?

examples/model/SDF_example_020.hs
SamuelMiksits, line 1: I already added the example from the labs as `examples/model/SDF_example_002.hs`, that file is exactly the same

src/CoreIRToProceduralIR.hs
klarasm: Right, you get the parameter list from the actor type? In that case it might be a little more work and I would ve fine keeping it this way.

examples/model/SDF_example_011.hs
sthaeron: Based on the lecture notes it should be netlist

examples/model/SDF_example_020.hs
sthaeron: Missed that, I'll drop the commit then.

src/CoreIR.hs
sthaeron: We were having problems with the Core IR pretty printer, so we decided to revert to what we originally had. Especially since all the other cases are using `showPpr` anyways.

src/ForSyDeIR.hs
sthaeron: It's just adding a type to the function instance for consistency with the rest of the code base.

src/CoreIRToProceduralIR.hs
sthaeron: I think we could potentially revisit this if we decide to redo how we handle errors. For now, this adds the programmer restriction that all functions must have an associated actor which calls them.

sthaeron:
@SamuelMiksits and @klarasm, I've added comments to the Core IR to Procedural IR translation solution and cleaned up some of the functions in the hopes that they are more readable.

SamuelMiksits (approved):
Yeah I think I am happy enough with this now

klarasm (approved):
Looks good!

PR #306, opened 2025-12-14
Allow delays connected to global inputs and outputs

### LINKED ISSUE #305: Allow for delays connected to global input/output in scheduler

klarasm:
Allow delays to be connected to global inputs and outputs.

There are a few parts of this:
1. SDFSchedule needs to find the correct rate when the IO edge is a delay. Do that by finding the first actor along that signal chain and use that rate.
2. This also does not generate a new IRId for delay edges and just uses the input one. Note that we still need the alias for the input signal since we always do the lookup.
3. ForSyDeIRToProceduralIR needs to use the translated buffers for `foldActorSignals` so it can see the actual buffer id.
4. Add examples for these scenarios.
5. Also remove the old hardcoded first attempt at translating Core functions to ProceduralIR.

### DIFF
--- a/examples/model/SDF_example_025.hs
+++ b/examples/model/SDF_example_025.hs
@@ -1,0 +1,29 @@
+module SDF_example_025 where
+
+import ForSyDe.Shallow
+
+-- Netlist
+system :: Signal Int -> Signal Int
+system s_in = s_out
+  where
+    (s_out_pre, first, second) = a_a s_in first_delayed second_delayed
+    first_delayed = d_1 first
+    second_delayed = d_2 second
+    s_out = d_out s_out_pre
+
+-- Process specifications
+a_a :: Signal Int -> Signal Int -> Signal Int -> (Signal Int, Signal Int, Signal Int)
+a_a s_1 s_2 s_3 = actor33SDF (1, 1, 1) (1, 1, 1) fib s_1 s_2 s_3
+
+d_1 :: Signal Int -> Signal Int
+d_1 s = delaySDF [0] s
+
+d_2 :: Signal Int -> Signal Int
+d_2 s = delaySDF [1] s
+
+d_out :: Signal Int -> Signal Int
+d_out s = delaySDF [0, 1] s
+
+-- Function definitions
+fib :: [Int] -> [Int] -> [Int] -> ([Int], [Int], [Int])
+fib [index] [first] [second] = ([index + first + second], [second], [index + first + second])
--- a/src/SDFSchedule.hs
+++ b/src/SDFSchedule.hs
@@ -85,14 +85,16 @@ convertIRSystem (IRSystem (inputNames, outputNames) constructors signals _) =
          in case (incoming, outgoing) of
               ([(inSignalId, srcIn, prodIn)], [(outSignalId, dstOut, consOut)]) ->
                 [ Edge
-                    (IRString $ show inSignalId ++ "_" ++ show outSignalId) -- delay edge names are combined
+                    -- use the input signal id as the delay edge id
+                    inSignalId
                     (findActorByName srcIn)
                     (findActorByName dstOut)
                     prodIn
                     consOut
                     True
                     (length delayConstructor)
-                    (Just [(inSignalId, IRString $ show inSignalId ++ "_" ++ show outSignalId), (outSignalId, IRString $ show inSignalId ++ "_" ++ show outSignalId)])
+                    -- allow the edge to be found by the output signal id as well
+                    (Just [(outSignalId, inSignalId)])
                 ]
               ([], _) ->
                 error $ "Delay node " ++ show delayName ++ " has no input signal."
--- a/src/SDFSchedule.hs
+++ b/src/SDFSchedule.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE LambdaCase #-}
 
 module SDFSchedule (computeScheduleAndBuffers, computeScheduleAndBuffersPrint) where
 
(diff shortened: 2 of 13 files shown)


src/SDFSchedule.hs
```diff
@@ -83,16 +84,23 @@ convertIRSystem (IRSystem (inputNames, outputNames) constructors signals _) =
...
+              -- input is a global system input -> ignore
+              ([(inSignalId, _, _)], _) | inSignalId `elem` inputNames -> []
+              -- output is a global system output -> ignore
+              (_, [(outSignalId, _, _)]) | outSignalId `elem` outputNames -> []
               ([(inSignalId, srcIn, prodIn)], [(outSignalId, dstOut, consOut)]) ->
                 [ Edge
-                    (IRString $ show inSignalId ++ "_" ++ show outSignalId) -- delay edge names are combined
+                    -- use the input signal id as the delay edge id
+                    inSignalId
                     (findActorByName srcIn)
```
Himawari722, line 95: This improvement looks great! I think you have written `nonDelayInputSignal` and `nonDelayInputSignal` to recursively find the first non delay signal, so it should support a delay actor chain and I wrote a simple example to test it. However, there was an error `Actor not found: a_delay2` when I tried running an SDF graph like this: 

`-- system: s_in -> a_add -> d1 -> d2 -> s_out`

I think the error was caused here `SDFSchedule:95`, when it was converting ForSyDe IR delay edges. It assumed the `src` and `dst` of an delay edge should be a non-delay actor so it just used `findActorByName` to find them, but it failed if there existed an delay chain. So there should also be a recursive way to build delay edges. Should I make a new issue and work on that?

klarasm: Right, we would need to merge successive delays between regular actors too (and ignore delay-delay connections if the chain is on an input or outout). I think it's more a nice to have since the same model with a single (merged) delay element on the edge should have the same semantic meaning (so we could get the same result by changing the model). Creating the issue would be good though, since we want to document what parts of the compiler is incomplete, and we can work on it if we have time left over.

SamuelMiksits (commented):
I think this is fine for the most part. I ran all of the models we have now and did not get any errors, and I tried the test suite and it passes. From the code perspective, I looked at it and it looks fine. But someone else might have other opinions or scenarios where it doesn't work

klarasm:
I added a few more examples to this PR.

examples/model/SDF_example_025.hs
```diff
@@ -0,0 +1,29 @@
+module SDF_example_018 where
```
sthaeron: Should be 25, `SDF_example_025`

sthaeron:
I'm a little confused about how the implementation works. I don't quite understand why adding `inputDelays` and `outputDelays` statement lists to the translation context was necessary. The signal names of system inputs and outputs, which are delayed, seem to be propagated to the process constructor for SDF examples 20 and 21. Thus, shouldn't the `foldActorSignal` pass for `a_add` add the necessary statements as they are global inputs/outputs? 

Perhaps we can talk about it tomorrow in person after the meeting @klarasm. But I can't seem to wrap my head around why the original solution didn't just work once the SDF Scheduler was updated?

Nevertheless, the implementation in its current state does work!

klarasm:
> I'm a little confused about how the implementation works. I don't quite understand why adding `inputDelays` and `outputDelays` statement lists to the translation context was necessary. The signal names of system inputs and outputs, which are delayed, seem to be propagated to the process constructor for SDF examples 20 and 21. Thus, shouldn't the `foldActorSignal` pass for `a_add` add the necessary statements as they are global inputs/outputs?
> 

~The reason the current solution doesn't work is that `foldActorSignals` only looks at the immediate input/output, meaning if there's a delay there, it won't produce the statements for reading or writing the buffer. But this actually gave me an idea, maybe it's better to make this function continue until it finds either an actor or global input/output (similar how I did it in the scheduler)?~

> Perhaps we can talk about it tomorrow in person after the meeting @klarasm. But I can't seem to wrap my head around why the original solution didn't just work once the SDF Scheduler was updated?
> 

We can do that.

> Nevertheless, the implementation in its current state does work!

klarasm:
@sthaeron

It was a good thing you commented on that, turns out the only thing we actually needed for `ForSyDeIRToProceduralIR` was:
```diff
diff --git a/src/ForSyDeIRToProceduralIR.hs b/src/ForSyDeIRToProceduralIR.hs
index f88c4091b8cc..43974dc59d88 100644
--- a/src/ForSyDeIRToProceduralIR.hs
+++ b/src/ForSyDeIRToProceduralIR.hs
@@ -193,8 +193,8 @@ translateIRConstructor initialContext constructor = case constructor of
                     ++ [EVar $ show functionId]
                 )
             )
-        inputStmts = foldl' foldActorSignals [] inputSignals
-        outputStmts = foldl' foldActorSignals [] outputSignals
+        inputStmts = foldl' foldActorSignals [] translatedInputSignals
+        outputStmts = foldl' foldActorSignals [] translatedOutputSignals
         stmts = reverse inputStmts ++ [actorCallStmt] ++ reverse outputStmts
         context1 = initialContext {actors = (actorId, stmts) : actors initialContext}
      in context1
```

Then the IO will be handled by the actors as you said.

klarasm:
Hmm. One issue with the approach in this PR is that even if we could produce output immediately we will still try to get input first. But this we can probably put as future work.

klarasm:
I reverted back to just placing parentheses around everything in this PR, though I have another idea in #309.

klarasm:
> Hmm. One issue with the approach in this PR is that even if we could produce output immediately we will still try to get input first. But this we can probably put as future work.

I think I address this one in the latest commit. I'm not sure about the formatting on the initial delay tokens on global output, though.

PR #317, opened 2025-12-21
docs: update some docs with what we have improved so far

Himawari722:


klarasm (approved):
I think this looks good!

PR #319, opened 2025-12-23
Undefined functions

sthaeron:
Adds support for undefined functions. Includes two additional SDF examples. In SDF example 31, the functions themselves are undefined, while in SDF example 32, undefined is used by the actors in place of a function name. Currently, only SDF example 31 is compilable to C code, while 32 fails to compile to ForSyDe IR. This is due to not having optional function IDs as a part of an `IRActor`, which it currently fails to find.

Also refines the approach used to ignore top-level Core IR constructs which relate to GHC-generated binders.

### DIFF
--- a/src/CoreIRToForSyDeIR.hs
+++ b/src/CoreIRToForSyDeIR.hs
@@ -5,6 +5,9 @@ import Data.List (elemIndex)
 import ForSyDeIR
 import GHC hiding (targetId)
 import GHC.Core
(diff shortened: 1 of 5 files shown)


src/CoreIRToForSyDeIR.hs
klarasm, line 476: Maybe we could just add IRString "undefined" for the case if we can't match a function (or maybe only when it's actually undefined)?

klarasm (approved):
Works on my computer!

PR #327, opened 2026-01-24
docs: add a user guide

klarasm:
Consolidate information on building and running the project. Also move the programmer restrictions into it and provide inline examples on what is and is not allowed.


sthaeron:
Ingo mentioned that using the commands `ghcup install ghc 9.10.2` and `ghcup set ghc 9.10.2` were unnecessary when using `stack`. However, I noticed that they are necessary when using `cabal`. If the wrong version of `ghc` is set, `cabal v1-install forsyde-shallow` will install the package globally for the wrong version of `ghc`. This means that the compiler and LSP won’t have access to ForSyDe Shallow. It does make sense that the user has to specifically install and use `ghc` 9.10.2, as they should be writing Haskell code for `ghc` 9.10.2, since that’s the version of `ghc` we are using and compiling.

sthaeron:
When following your visualizer installation instructions, I ran into a few issues. There is no mention in the instructions of installing the KLighD Diagrams extension, which is a dependency of our visualizer. I also noticed when testing in a VM with a fresh install of VSCode, our visualizer also requires the Haskell extension. This is most likely due to the Haskell extension configuring `.hs` files to be detected by VSCode as Haskell files.

When I attempted to install and use `vsce` to package and install our visualizer extension, I ran into some user permission errors. I tried fixing the problem using the info found in [npm docs](https://docs.npmjs.com/resolving-eacces-permissions-errors-when-installing-packages-globally). But this did not work for me. Using the flag `--extensionDevelopmentPath=` did work.

We should also add instructions relating to enabling `Client Layout` found under `Debug Options` for the visualizer diagrams to start rendering.

klarasm:
> Ingo mentioned that using the commands `ghc install ghc 9.10.2` and `ghc set ghc 9.10.2` were unnecessary when using `stack`. However, I noticed that they are necessary when using `cabal`. If the wrong version of `ghc` is set, `cabal v1-install forsyde-shallow` will install the package globally for the wrong version of `ghc`. This means that the compiler and LSP won’t have access to ForSyDe Shallow. It does make sense that the user has to specifically install and use `ghc` 9.10.2, as they should be writing Haskell code for `ghc` 9.10.2, since that’s the version of `ghc` we are using and compiling.

Do you mean `ghcup install ...` and `ghcup set ...`? But it makes sense that it did not work for me since my system ghc is 9.8.4.

sthaeron:
I'm currently writing an updated instruction for the compiler and LSP, perhaps you could work on updating the visualizer instructions?

klarasm:
> I'm currently writing an updated instruction for the compiler and LSP, perhaps you could work on updating the visualizer instructions?

Yes, also currently addressing your comments.

klarasm:
(I still need to add notes about the client layout)

klarasm:
@sthaeron I think that should be everything you commented on.

klarasm:
> I'm currently writing an updated instruction for the compiler and LSP, perhaps you could work on updating the visualizer instructions?

Or do you mean something separate, like a quick start?

sthaeron:
I've added a new commit which moves the installation instructions to the project README. I've also moved your visualizer instructions, but did not make any changes. I'll let you make any changes you see fit. I've also reorganized the usage and programmer restriction sections in the user guide to follow the same order and the installation instructions.

README.md
```diff
@@ -1,28 +1,89 @@
...
+The visualiser depends on [KLighD
+Diagrams](https://marketplace.visualstudio.com/items?itemName=kieler.klighd-vscode)
+as well as
+[Haskell](https://marketplace.visualstudio.com/items?itemName=haskell.haskell)
+VS Code extensions which need to be installed independently.
+In addition, you also need
+[npm](https://nodejs.org/en/learn/getting-started/an-introduction-to-the-npm-package-manager)
+for building the extension.
+
+To build and install the VSCode extension:
```
sthaeron, line 61: Should probably mention that the VSCode extension requires the ForSyDe DevTools LSP to be properly installed and available from your shell path. Since someone might skip directly to the extension installation, not realizing they needed to follow the previous instructions.

docs/user-guide.md
```diff
@@ -0,0 +1,224 @@
...
+CTRL+Shift+P and select "Developer: Reload Window" (start to write reload
+window in the text field if it does not appear).
+
+If you have no diagram (or an outdated one) or the "Current synthesis:"
+drop-down is empty, try the above step. For further debugging you can select
+"ForSyDe DevTools LSP" in the Output window. E.g. in the case when the language
+server complains about not finding ForSyDe Shallow when importing it, recheck
+the build instructions (likely for stack).
+
+## Using the language server separately
```
sthaeron, line 95: Missing LSP usage section introduction sentence. Should add:
>  The output of the usage and help prompts for the ForSyDe DevTools language server is shown below for reference:

README.md
```diff
@@ -1,28 +1,89 @@
...
+```sh
+cd ./vscode-ext
+npm install
+npm run compile
+./node_modules/vsce/vsce package
```
sthaeron, line 66: Tried the updated visualizer installation instructions and it worked great for me! Its much simpler than before.

sthaeron (commented):
Just added a few comments. The new visualizer instructions worked great for me. Otherwise, everything else looks good to me.

klarasm:
Thanks, updated now.

sthaeron (approved):
Looks good to me, feel free to merge the PR!

PR #339, opened 2026-02-25
update copyright with all the contributors

### LINKED ISSUE #340: Change the license
As per conversation with Ingo, the license should not just refer to Sebastian, but the entire team, and maybe ForSyDe lab as well? We should confirm this with the Ingo in that case.

The license is also optionally required in the VS Code Extension, so it should also be copied to the `vscode-ext` dir.

klarasm:


klarasm:
@sthaeron this now contains the other things we discussed.

sthaeron (approved):
Looks good to me!

klarasm:
@ramadhanafif does this also look good to you? In that case I'll merge it.

ramadhanafif:
✅ good

PR #341, opened 2026-03-04
refactor(vscode-ext): clean up unused or incorrect configuration

ramadhanafif:
clean dead codes, unused file, and make the overall vsix bundle contains metadata that can be consumed by the VS COode.

PR #342, opened 2026-03-05
Add CI Workflow for Github Release

ramadhanafif:
Add CI Workflow that create github release when we push a tag that starts with v, such as v123456, or v1.2.3.4 or even tag named "vaxholm". See image below.

The release contains two important outcomes:
1. Compressed file `something-something.tar.gz` which bundles the `forsyde-lsp-exe` and `forsyde-compiler-exe` for Linux. (windows and mac later)
2. VS Code extension in `.vsix` format 

image

Note: I accidentally rebased on a unmerged PR, so do not merge before PR #341 is merged.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,101 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - 'v*'
+
+jobs:
+  build-haskell:
+    name: Build Haskell Binaries
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Install Nix
+        uses: cachix/install-nix-action@v31
+        with:
+          nix_path: nixpkgs=channel:25.05
+
+      - name: Build binaries
+        run: |
+          nix develop --command bash -c "cabal update && cabal build all"
+
+      - name: Package binaries
+        run: |
+          mkdir -p dist/bin
+          # Find and copy the executables produced by cabal
+          # They are usually deeply nested in dist-newstyle, so we use find
+          find dist-newstyle -type f -executable -name "forsyde-compiler-exe" -exec cp {} dist/bin/ \;
+          find dist-newstyle -type f -executable -name "forsyde-lsp-exe" -exec cp {} dist/bin/ \;
+          
+          # Create an archive
+          cd dist
+          tar -czvf forsyde-devtools-linux-x86_64.tar.gz bin/
+
+      - name: Upload Haskell artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: haskell-binaries
+          path: dist/forsyde-devtools-linux-x86_64.tar.gz
+          retention-days: 1
+
+  build-vscode-ext:
+    name: Build VS Code Extension
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '22'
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,101 @@
...
+      - name: Install Nix
+        uses: cachix/install-nix-action@v31
+        with:
+          nix_path: nixpkgs=channel:25.05
+
+      - name: Build binaries
+        run: |
+          nix develop --command bash -c "cabal update && cabal build all"
+
+      - name: Package binaries
```
klarasm, line 40: The nix binaries won't work if nix is not installed on the host computer? (Correct me if I'm wrong @sthaeron)

klarasm (commented):
I'm not sure we should use the nix built binaries. The stack binaries might be better (then the use can only install BLAS and LAPACK). The automatic packaging of the VS Code extension is really nice though.

klarasm:
> 
> Note: I accidentally rebased on a unmerged PR, so do not merge before PR #341 is merged.

You should be able to resolve this with something like `git rebase vscode-ext-cleanup --onto upstream/dev`.

ramadhanafif:
> > Note: I accidentally rebased on a unmerged PR, so do not merge before PR #341 is merged.
> 
> You should be able to resolve this with something like `git rebase vscode-ext-cleanup --onto upstream/dev`.

Thanks for the suggestion! It happen that this PR actually using some of the new build command for the vs code extension, such as `npm run package`, so there is an actual dependency to the other commit.

.github/workflows/release.yml
ramadhanafif: Can confirm that this does not work.

ramadhanafif:
Revert back this PR to draft. 
TODO: change forsyde binaries from nix build to depend on stack instead.

ramadhanafif:
The latest commit works. You can download it in [this release on my fork](https://github.com/ramadhanafif/forsyde-devtools/releases/tag/v123),

then extract it with 
`tar xvf forsyde*.tar.gz`

finally test it with docker one-liner:
`docker run --rm -it -v (pwd)/bin/forsyde-compiler-exe:/usr/local/bin/forsyde-compiler-exe debian:13 bash -c "apt update && apt install -y libblas3 liblapack3 && forsyde-compiler-exe"`

put $ in the pwd if you are using bash.

klarasm:
Hmm:
```sh
$ ./bin/forsyde-compiler-exe ~/git/forsyde-devtools/examples/model/SDF_example_002.hs --stdout
forsyde-compiler-exe: Missing file: /home/runner/.stack/programs/x86_64-linux/ghc-tinfo6-9.10.2/lib/ghc-9.10.2/lib/settings
HasCallStack backtrace:
  finally, called at compiler/GHC/Utils/Panic.hs:303:7 in ghc-9.10.2-2f07:GHC.Utils.Panic
```

ramadhanafif:
> Hmm:
> ```sh
> $ ./bin/forsyde-compiler-exe ~/git/forsyde-devtools/examples/model/SDF_example_002.hs --stdout
> forsyde-compiler-exe: Missing file: /home/runner/.stack/programs/x86_64-linux/ghc-tinfo6-9.10.2/lib/ghc-9.10.2/lib/settings
> HasCallStack backtrace:
>   finally, called at compiler/GHC/Utils/Panic.hs:303:7 in ghc-9.10.2-2f07:GHC.Utils.Panic
> ```

I think this might be your linux setup problem...
Can you do `ldd forsyde-compiler-exe`? 

Edit: it panics on my machine as well...

ramadhanafif:
Good old problem of "it worked on my machine"

PR #343, opened 2026-03-10
Start working on error handling

klarasm:
- Error handling. This is primarily important for the LSP so it doesn't crash if there is some intermediary invalid models. These files are converted to `Either String` instead of `error`
    1. SDFSchedule
    2. CoreIRToForSyDeIR
- Some cleanup, like defining the type for the schedule in the scheduler and trimming export lists
- Fix a bug regarding delays on system input
- Add a simple test scripts which can handle systems of a single input and output and run it in the CI

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -1,0 +1,59 @@
+#!/bin/bash
+tests=$(find examples/model examples/atom -name '*.hs' -type f -exec grep -l '^main' {} +)
+
+inputs=12
+ret=0
+
+if command -v stack; then
+	build_hs="stack ghc --"
+	run_comp="stack run forsyde-compiler-exe --"
+else
+	build_hs="cabal exec ghc --"
+	run_comp="cabal exec forsyde-compiler-exe --"
+fi
+
+while [ -n "$1" ]; do
+	case "$1" in
+		"--stack")
+			build_hs="stack ghc --"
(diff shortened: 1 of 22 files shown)


klarasm:
@sthaeron this should be ready for review now.

scripts/test.sh
```diff
@@ -0,0 +1,59 @@
...
+	gcc -I examples/implementation/platform_independent main.c
+
+	echo "Running $name:"
+	input=$( ( for i in $(seq $inputs); do echo -n " $(( RANDOM % 100 ))"; done; ) )
+	echo "Input:$input"
+	echo "$input" | ./$exe > "$output_hs"
+	echo "$input" | ./a.out | { tr '\n' ' '; echo; } | sed -e 's/  */ /g' -e 's/ *$//' > "$output_c"
+
+	if ! diff -Naur "$output_hs" "$output_c"; then
+		echo "Failed $name"
```
sthaeron: Should probably be `echo "$name: FAILED"` to stay constant with the success message on line 51.

sthaeron (approved):
All changes look great! I’ve tested the new test script locally and had to make some changes to the nix environment so that I do not use my local gcc instance. I’ve also tested the VSCode extension and managed to get the new scheduler error in the output log and the LSP did not crash, great work!

I’ve added an additional commit to your pr that just updates the nix flake adding gcc.

PR #344, opened 2026-03-25
Add support for ForSyDe-Atom SDF

klarasm:
The only thing needed was to add new patterns for `translateCoreExpr` in `CoreIRToForSyDeIR.hs` and adding the dependencies.

Note: I don't know how to add ForSyDe-Atom to the nix flake since it is not on hackage.

### DIFF
--- a/examples/atom/SDF_example_001.hs
+++ b/examples/atom/SDF_example_001.hs
@@ -1,0 +1,27 @@
+module SDF_example_001 where
+
+import qualified ForSyDe.Atom.MoC.SDF as SDF
+
+system :: SDF.Signal Int -> SDF.Signal Int -> SDF.Signal Int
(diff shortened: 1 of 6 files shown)


sthaeron:
That's great to hear, I'll had ForSyDe-Atom to the nix flake and test it on my system.

sthaeron:
I've added some commits to your branch. One which update's the nix flake to include ForSyDe-Atom.  The other adds an example of using ForSyDe-Atom. I've tested your changes on this example and it worked translating to C, compiles, and output is correct.

klarasm:
@sthaeron I think your changes look good, nice to have the example. Do you think I should do anything more before merging?

sthaeron:
It might be worth it to test more examples, perhaps their's some edge case using eta reduction.

PR #345, opened 2026-03-25
vscode-ext: invoke the lsp through stack

klarasm:
This means that the user does not have to specify the --forsyde-pkgpath, though I'll leave it since it should not have any ill effect.

NOTE: this should probably not be merged before #344.


klarasm:
Hmm. This has some issues when forsyde-atom is not in the stack.yaml (since forsyde-atom is not on hackage).

klarasm:
Provide a `stack.yaml` instead which can specify the git repo for forsyde-atom

PR #347, opened 2026-04-11
Update install

sthaeron:
I've started updating the install instructions and user-guide based on that we now recommend running the compiler and LSP using stack. I've also updated the VSCode extension instructions, specifying the version of `npm` we recommend.


sthaeron:
@klarasm is there anything else we should update in the instruction or user-guide?

README.md
```diff
@@ -53,7 +47,7 @@ First, install the compiler and LSP as described in the previous section.
 Install the VSCode extensions [KLighD
 Diagrams](https://marketplace.visualstudio.com/items?itemName=kieler.klighd-vscode)
 and
-[Haskell](https://marketplace.visualstudio.com/items?itemName=haskell.haskell).
+[Haskell](https://marketplace.visualstudio.com/items?itemName=haskell.haskell). You will also need `npm`, at least version `11.12.*`, to build and package the extension. We recommend installing `npm` through the following [link](https://nodejs.org/en/download).
 In addition, you also need
```
klarasm: This sentence is now redundant with the bit you add above.

README.md
```diff
@@ -63,7 +57,7 @@ To build and install the VSCode extension:
 cd ./vscode-ext
 npm install
 npm run compile
-./node_modules/vsce/vsce package
+./node_modules/@vscode/vsce/vsce package
```
klarasm: This should be `npm run package` (@ramadhanafif added this in #341).

stack.yaml
klarasm, line 1: Nice.

README.md
klarasm, line 18: Do we need to tell them to clone the repo here? I don't think cloning the repo before this point would be an issue.

docs/user-guide.md
klarasm: Is this still true? I guess it still wouldn't hurt to keep the general recommendation, though.

klarasm: Did your change allow for arbitrary number of inputs/outputs?

klarasm: Unrelated, but I have a rough idea for how we could support if expressions. There is a GNU C extension which is widely supported which allows making an expression of any statement, e.g.:
```C
int statement_as_expression(int a, int b)
{
    return ({
        int c;
        if (a > b)
            c = a*b;
        else
            c = a+b;
        c;
    });
}
```

klarasm: While this is likely now a lot less common I think it makes sens to keep it. The case for removing it would be to keep it simpler. What do you think?

klarasm, line 55: Is it an issue that the ES-Lab-Kit repo is only accessible to people with a KTH email?

klarasm (changes requested):
I have some comments, also added thoughts on more things to improve.

README.md
sthaeron: I'll remove the `ghcup` setting and install version `9.10.2` instructions. Since, they just need to install `stack` to build and run the project now, which itself handles the `ghc` versioning.

This will bring the git cloning instructions closer to the top.

docs/user-guide.md
sthaeron: Forgot to update the user guide, will do that.

sthaeron: Do we really need it to be all one expression in C to support if statements. Can't we just have a return at the end of each scope?

sthaeron: Perhaps we could ask Matthias or Ingo if the lab kit repo can be published on the public github.

klarasm: A return will always return from the function, not the scope so that won't work in the general case (if that's what you mean?). However, we probably don't need this either. I first thought it would be easier to assign array members this way but we could just assign it the value of `c` for this example at least. For most cases a ternary would probably be enough anyhow.

README.md
klarasm: I think you have to configure stack to use ghcup for fetching ghc. By default it uses its own I think, but maybe that would work to.

sthaeron: When you install `ghcup` is asks whether it should configure `stack` to use `ghcup` to install different versions of `ghc`. Nevertheless, we don't really care how they get or setup `stack`. We just need them to have a working install of `stack`.

klarasm: Right, I probably didn't see that since I used my distribution's version of `ghcup` and `stack`.

But this reminded me that we should probably state which version of HLS the user should have for 9.10.2. I think it was 2.11.0.0?

docs/user-guide.md
sthaeron: I think this is something we should discuss in person. I seem to be misunderstanding your reasoning and the problem.

sthaeron:
@klarasm, I've updated the install instructions of the README and updated the user-guide based on your suggestions.

klarasm (approved):
I think this looks good now!

Issue #292, opened 2025-12-03
Look at alternatives to hmatrix


klarasm:
`new chaos`

### What to change
Remove non-default C library dependencies for our project.

### How to change
Hmatrix depends on opanblas and lapack which are C libraries which might not be installed on the user's computer.

Check if it's feasible to use a pure haskell matrix library, perhaps https://hackage.haskell.org/package/matrices or something else. (This might also allow us to use rational numbers).

### Why to change
This would allow us to only have haskell dependencies which would make the executable easier to statically link, as well as being able to build with cabal without installing external dependencies.

### When to change
When we need to package our project.

sthaeron:
This would be a great change, especially now that we no longer have `clang-format` as an external dependency. It'll ensure that users will only need `cabal` to install our compiler and lsp from the Hackage Haskell package repository. We can also expect our users to have `cabal` or `stack` working, since they'll need it to set up ForSyDe Shallow.

klarasm:
(probably not https://hackage.haskell.org/package/bed-and-breakfast is pretty bare)

(nix does not like https://hackage.haskell.org/package/HaskellForMaths-0.4.9/docs/Math-Algebra-LinearAlgebra.html but it seems to have what we need)

SamuelMiksits:
I tried the `stack` PR and honestly installing `blas-devel` and `lapack-devel` isn't too bad and not far from just general types of project that you build. Looking at what we are using from the current library it is rank (`rank` in hmatrix) and null (`nullspace` in hmatrix). 

`bed-and-breakfast` has rank but not null from what I can tell.

Maybe I am being a smartass here but if all we need is those 2 (please fill in if I am missing something) then it can't be that hard to handroll our own library that does those 2 functions. In my C++ course we wrote our own C++ library for a Matrix datatype so when I have been learning other languages I have been implementing my own Matrix library for fun, so about a year and a half ago I wrote a "flat array" Matrix class that I can share if you are interested. I implemented matrix-matrix add and matrix-matrix multiplication. In Main you can declare and use them like:

```hs
main :: IO ()
main = do

    let m1 = Matrix 2 2 [1::Int,2,3,4]
    let m2 = Matrix 2 2 [3::Int,4,5,6]
    let m3 = m1 + m2
    print m3
    
    let m4 = m1 * m2
    print m4
```

(so you pass a flat array to initialize the matrix and then you have to say how many rows and columns the matrix has) which outputs
```
Matrix: (2,2):
[...]

PR #348, opened 2026-07-17
refactor: replace hmatrix with exact rational arithmetic in scheduler

ramadhanafif:
Removes the OpenBLAS/LAPACK system dependency (#292). The repetition vector is now computed with exact Gaussian elimination over rationals, replacing the floating-point nullspace and its tolerance-based integer normalization. Null space verification is exact instead of using a 1e-9 threshold, and the schedule printer shows exact fractions.

This commit is made by AI because i'm not fancy enough to deal with matrices in haskell. it passes tests but it might be worth putting extra care in checking the commits. Reminding all again that the motivation in doing this is so we can ship a vscode extension with no system dependency.

Solves #292


klarasm:
Nice! I'll take a look at some point.

klarasm (approved):
I think this looks good! Really nice to have exact fractions.

Issue #133, opened 2025-10-07
Look at packaging the result properly


klarasm:
Do we ship our own ghc? Or something else?

PR #349, opened 2026-07-17
add release pipeline with packaged vscode extension

### LINKED ISSUE #280: Prototype on how to bundle and publish the VSCode Extension.
Output of this issue is a .vsix file that can be shared and installed in other's PC.

ramadhanafif:
pushing a version tag (v0.1.0 etc) now produces a github release with an installable .vsix that has the LSP server bundled inside, plus a tarball of the standalone binaries. closes #280, progress on #133.

changes:
- new .github/workflows/release.yml, runs on v* tags and manual dispatch. stack build (cached), copies forsyde-lsp-exe into vscode-ext/server/, packages with vsce, attaches the vsix and forsyde-devtools-linux-x64.tar.gz to the release
- extension.ts: the bundled server was being launched with the stack exec args, which the binary can't parse. bundled binary now gets the lsp args directly
- README: added "installing from a release" section before the build-from-source instructions
- rebased onto dev after #348, so blas/lapack no longer needed. dropped the install step from the workflow and the blas note from the readme

linux x64 only for now, workflow can grow a matrix later. users still need stack + a compiled forsyde-shallow at runtime, the release just removes the npm/vsce/clone steps from install.

not tested end to end yet, needs a tag push or manual dispatch. maybe a v0.1.0-rc1 dispatch after merge.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -23,12 +23,6 @@ jobs:
     steps:
       - uses: actions/checkout@v5
 
-      # The compiler requires the OpenBLAS and LAPACK external libraries
-      - name: Install system dependencies
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y libblas-dev liblapack-dev
-
       - name: Set up stack
         uses: haskell-actions/setup@v2
         with:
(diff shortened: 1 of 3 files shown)


ramadhanafif:
once the dependent PR is merged then i will mark this as not draft.

klarasm (commented):
The binaries still don't work on my computer:
```
forsyde-lsp-exe: Missing file: /home/runner/.stack/programs/x86_64-linux/ghc-tinfo6-9.10.2/lib/ghc-9.10.2/lib/settings
HasCallStack backtrace:
  finally, called at compiler/GHC/Utils/Panic.hs:303:7 in ghc-9.10.2-2f07:GHC.Utils.Panic
```

